PR #1, opened 2021-07-14
Initial setup phase-1

Baggerone:


application/api/errorcodes.go
```diff
@@ -0,0 +1,42 @@
...
+	ErrorNoRows                   = ErrorKey("ErrorNoRows")
+	ErrorNotAuthorized            = ErrorKey("ErrorNotAuthorized")
+	ErrorQueryFailure             = ErrorKey("ErrorQueryFailure")
+	ErrorSaveFailure              = ErrorKey("ErrorSaveFailure")
+	ErrorTransactionNotFound      = ErrorKey("ErrorTransactionNotFound")
+	ErrorUnknown                  = ErrorKey("ErrorUnknown")
+	ErrorUpdateFailure            = ErrorKey("ErrorUpdateFailure")
+	ErrorValidation               = ErrorKey("ErrorValidation")
+
+	// HTTP codes for customErrorHandler
```
Schparky, line 28: This section will not be needed. As I recall, `customErrorHandler` was something we needed for gqlgen

application/locales/global.en.yaml
```diff
@@ -0,0 +1,4 @@
+
+# User
+- id: GetUser
+  translation: We had a problem finding the user profile.
```
Schparky, line 4: minor detail: can we establish these defaults in GoLand (and the equivalent in VSCode)?
image

application/api/api.go
```diff
@@ -0,0 +1,141 @@
...
+// keyToReadableString takes a key like ErrorSomethingSomethingOther and returns Error something something other
+func keyToReadableString(key string) string {
+	re := regexp.MustCompile(`[A-Z][^A-Z]*`)
+	words := re.FindAllString(key, -1)
+
+	if len(words) == 0 {
+		return key
+	}
+
+	if len(words) == 1 {
```
jason-jackson, line 111: Do we want this to be to lower too? (And is it needed?)

application/api/errorcodes.go
```diff
@@ -0,0 +1,42 @@
+package api
+
+const (
+	CategoryDatabase  = ErrorCategory("DB")
```
jason-jackson, line 4: "DB" doesn't seem to follow names of other categories, would "Database" be better?

application/models/models.go
```diff
@@ -0,0 +1,55 @@
...
+	if err != nil {
+		return "", err
+	}
+
+	return base64.URLEncoding.EncodeToString(rb), nil
+}
+
+
+
+// CurrentUser retrieves the current user from the context, which can be the context provided by the inner
```
Schparky, line 49: I think when you trimmed the gqlgen stuff, you left a few details in the function doc that no longer make sense

application/models/user.go
```diff
@@ -0,0 +1,16 @@
...
+import (
+	"time"
+
+	"github.com/gofrs/uuid"
+)
+
+// User model
+type User struct {
+	ID        int       `json:"-" db:"id"`
+	CreatedAt time.Time `json:"created_at" db:"created_at"`
```
Schparky, line 12: If we're going to use an `api` package, I believe the model no longer needs `db` tags.

jason-jackson (approved):

Schparky (approved):

.gitignore
```diff
@@ -11,5 +11,27 @@
 # Output of the go coverage tool, specifically when used with LiteIDE
 *.out
 
-# Dependency directories (remove the comment below to include it)
-# vendor/
+
+application/swagger.json
```
jason-jackson, line 15: Could we sort these?

application/api/api.go
```diff
@@ -0,0 +1,141 @@
...
+		return words[0]
+	}
+
+	for i := 1; i < len(words); i++ {
```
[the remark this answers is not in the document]
Baggerone: The loop doesn't call ToLower on the first word.

application/api/api.go
Baggerone: :+1:  It's not needed, since the loop will be skipped if len(words) == 1

application/api/errorcodes.go
Baggerone: OK

application/api/errorcodes.go
Baggerone: :+1:

application/locales/global.en.yaml
fillup: Where is that?

application/models/user.go
fillup: so long as the property name maps to the db column with the default mapper right?

application/locales/global.en.yaml
Baggerone: :+1:

application/models/models.go
Baggerone: :+1:

.gitignore
Baggerone: :+1:

application/locales/global.en.yaml
Baggerone: In Goland, it's near the bottom of Settings > General

application/models/user.go
Schparky: I meant to say "no longer needs json tags"

fillup (approved):

PR #2, opened 2021-07-14
Initial setup for tests

Baggerone:


.env.example
```diff
@@ -0,0 +1,12 @@
...
+AWS_SECRET_ACCESS_KEY=abcd1234
+
+DISABLE_TLS=true
+
+RISKMAN_API_BASE_URL=localhost
+RISKMAN_APP_NAME=riskman
+RISKMAN_GO_ENV=development
+
+RISKMAN_ROLLBAR_SERVER_ROOT=github.com/myorg/myapp
+RISKMAN_ROLLBAR_TOKEN
```
Schparky, line 12:
```suggestion
RISKMAN_ROLLBAR_TOKEN=
```

application/Dockerfile-dev
```diff
@@ -0,0 +1,39 @@
...
+
+# Copy the Go Modules manifests
+# cache deps before building and copying source so that we don't need to re-download as much
+# and so that source changes don't invalidate our downloaded layer
+COPY go.mod go.mod
+COPY go.sum go.sum
+RUN go mod download
+
+# additional deps needed for running tests
+RUN go get github.com/gobuffalo/suite/v3
```
Schparky, line 29: I think this can be all on one line.
```suggestion
RUN go get github.com/gobuffalo/suite/v3 github.com/gobuffalo/httptest github.com/markbates/grift github.com/stretchr/testify github.com/gorilla/pat github.com/gorilla/context
```

application/models/testutils_test.go
```diff
@@ -0,0 +1,50 @@
...
+	"reflect"
+	"strconv"
+
+	"github.com/gobuffalo/pop/v5"
+
+	"github.com/silinternational/riskman-api/domain"
+)
+
+// mustCreate saves a record to the database. Panics if any error occurs.
+func mustCreate(tx *pop.Connection, f interface{}) {
```
Schparky, line 14: Do we want to make test fixture functions exportable so we don't have to make two copies? i.e. moved to a file without "_test" and capitalized?

Schparky (approved):

application/domain/domain.go
```diff
@@ -11,6 +11,7 @@ import (
 	"github.com/gobuffalo/buffalo"
 	mwi18n "github.com/gobuffalo/mw-i18n"
 	"github.com/gobuffalo/packr/v2"
+	uuid2 "github.com/gofrs/uuid"
```
fillup, line 14: why the alias?

application/models/testutils_test.go
fillup: worth a try. this could even be a case to use a build tag to only have it included when `GO_ENV` is `test` or something so we're never bundling the test helper stuff in the deployed app.

fillup: Maybe it cant work with existing env vars, but we can pass tags to `go build -tags xyz` to include or not include some files. Check out https://www.digitalocean.com/community/tutorials/customizing-go-binaries-with-build-tags

docker-compose.yml
```diff
@@ -0,0 +1,106 @@
+version: "3.7"
+
+services:
+  proxy:
+    image: silintl/traefik-https-proxy
```
fillup, line 5: Lets drop the traefik proxy in this case. I don't think we need ssl for anything. if we need it later we'll bring it back in

fillup (approved):
I added a few comments, but otherwise approved.

.env.example
Baggerone: :+1:

application/domain/domain.go
Baggerone: Looks like we can remove it.  I'm guessing we just had a `uuid` variable in the code which resulted in a naming collision.

PR #5, opened 2021-07-15
Add simple grifts and actions

Baggerone:


fillup (approved):

application/domain/domain.go
```diff
@@ -154,3 +156,27 @@ func GetUUID() uuid.UUID {
 	}
 	return id
 }
+
+func RollbarMiddleware(next buffalo.Handler) buffalo.Handler {
+	return func(c buffalo.Context) error {
+		if Env.RollbarToken == "" {
```
jason-jackson, line 162: It's somewhat tempting to combine the two if statements. It's also somewhat tempting to just put the middle (set client and add to context) in`if Env.RollbarToken != "" && Env.GoEnv != test {` instead since the method ends with the `return next(c)`. But I'm fine with any way.

jason-jackson (approved):

Schparky (approved):

PR #6, opened 2021-07-16
try different validation library

Baggerone:


jason-jackson (approved):

fillup (approved):

application/models/user.go
```diff
@@ -14,6 +18,16 @@ type User struct {
 	ID        uuid.UUID `json:"-" db:"id"`
 	CreatedAt time.Time `db:"created_at"`
 	UpdatedAt time.Time `db:"updated_at"`
-	Email     string    `db:"email"`
-	UUID      uuid.UUID `db:"uuid"`
+	Email     string    `db:"email" validate:"required"`
+	UUID      uuid.UUID `db:"uuid" validate:"required"`
```
Schparky, line 22: Guess I missed this earlier, but why are there two UUID fields on User?

Schparky (approved):

application/models/user.go
Baggerone: Should I make the `ID` field an `int` or get rid of the `UUID` field?

PR #8, opened 2021-07-31
Initial AuthN and AuthZ middlewares

fillup:


application/actions/users.go
```diff
@@ -4,8 +4,26 @@ import (
...
+
+func usersView(c buffalo.Context) error {
+	if user := getUserFromCxt(c); user != nil {
+		return c.Render(http.StatusOK, r.JSON(user))
+	}
+	return c.Render(http.StatusNotFound, nil)
+}
+
+func getUserFromCxt(c buffalo.Context) *models.User {
+	user, ok := c.Value(domain.TypeUser).(models.User)
```
Baggerone, line 24: Instead of `TypeUser`, should this be `ContextKeyCurrentUser`?

Baggerone (approved):
One thing to change, unless I'm wrong about it.

PR #14, opened 2021-08-03
authz for policy, policy users, and policy dependents

fillup:


application/models/policy.go
```diff
@@ -47,3 +51,33 @@ func (p *Policy) GetID() uuid.UUID {
...
 	return tx.Find(p, id)
 }
+
+// IsActorAllowedTo ensure the actor is either an admin, or a member of this policy to perform any permission
+func (p *Policy) IsActorAllowedTo(tx *pop.Connection, user User, perm Permission, sub SubResource, r *http.Request) bool {
+	if user.IsAdmin() {
+		return true
+	}
+
+	if len(p.Members) == 0 {
```
Baggerone, line 61: I'm wondering if it's better to have something like this in the LoadMembers method, instead of repeating the check in the calling functions.  `if len(p.Members) > 0 {return}`

Baggerone (approved):

Schparky (approved):

application/models/policy.go
fillup: oh yeah, smart!

PR #16, opened 2021-08-03
items

fillup:


application/migrations/20210803202823_create_items_tables.up.fizz
```diff
@@ -0,0 +1,23 @@
...
+	t.Column("country", "string", {})
+	t.Column("description", "string", {"size": 2048})
+	t.Column("policy_id", "uuid", {})
+	t.Column("policy_dependent_id", "uuid", {"null": true})
+	t.Column("make", "string", {})
+	t.Column("model", "string", {})
+	t.Column("serial_number", "string", {})
+	t.Column("coverage_amount", "int", {})
+	t.Column("purchase_date", "datetime", {})
+	t.Column("coverage_status", "string", {})
```
jason-jackson, line 15: do we want to do something like: 
`sql("CREATE TYPE status AS ENUM ('draft','pending','approved','denied');")`
`t.Column("coverage_status", "status", {})`

Baggerone (approved):

application/migrations/20210803202823_create_items_tables.up.fizz
```diff
@@ -0,0 +1,23 @@
...
+	t.Column("coverage_start_date", "datetime", {})
+	t.Timestamps()
+
+	t.ForeignKey("category_id", {"item_categories": ["id"]}, {})
+	t.ForeignKey("policy_id", {"policies": ["id"]}, {"on_delete": "cascade"})
+	t.ForeignKey("policy_dependent_id", {"policy_dependents": ["id"]}, {"on_delete": "SET NULL"})
+	t.Index(["name", "policy_id"], {"unique": true})
```
Schparky, line 22: Doesn't this lock in a 1-to-1 on `policies` and `items`? OK for this week, but don't we want multi-item policies?

application/migrations/20210803202823_create_items_tables.up.fizz
jason-jackson: I can understand if we don't want to limit ourselves right now to a specific set of statuses.

Schparky (approved):

application/migrations/20210803202823_create_items_tables.up.fizz
fillup: thanks for the suggestion, but we like to manage the options as constants so that we don't have to do migrations anytime we change the options

application/migrations/20210803202823_create_items_tables.up.fizz
fillup: this ensures the name is unique in a policy. so you cant have two items with identical names. maybe that isn't worth enforcing, but it seemed like a good idea to prevent confusion, especially if they have different coverage amounts

PR #18, opened 2021-08-04
start on policy endpoints and api types

fillup:


application/actions/policies.go
```diff
@@ -0,0 +1,49 @@
...
+	"net/http"
+
+	"github.com/gobuffalo/buffalo"
+	"github.com/silinternational/riskman-api/models"
+)
+
+func policiesList(c buffalo.Context) error {
+	user := getUserFromCxt(c)
+
+	if user.IsAdmin() {
```
jason-jackson, line 13: Edge case: What should happen if admin wants to view their policies?

Baggerone (approved):

application/actions/policies.go
```diff
@@ -0,0 +1,49 @@
...
+
+import (
+	"net/http"
+
+	"github.com/gobuffalo/buffalo"
+	"github.com/silinternational/riskman-api/models"
+)
+
+func policiesList(c buffalo.Context) error {
+	user := getUserFromCxt(c)
```
Schparky, line 11: I guess I missed this from an earlier PR, but we have used `CurrentUser()` as the name for this function. (Or, if keeping this name, I think context is usually abbreviated "Ctx".)

Baggerone:
Looks like  you created an import cycle.
I'll have a look.

jason-jackson (approved):

Baggerone:
I put the PolicyTypes in api to fix the import cycle.

Schparky (approved):

application/actions/policies.go
fillup: good catch! so far we've confirmed the only admin does not have a policy and is not currently eligible to have one anyway

application/actions/policies.go
fillup: actually I think this is an error on my part. this method pulls the user referenced in an api call like /users/{id}, not the authenticated user. I'll rename this method to getReferencedUser... to help that be less confusing

application/actions/policies_test.go
```diff
@@ -0,0 +1,98 @@
...
+		NumberOfPolicies:    3,
+		ItemsPerPolicy:      2,
+		UsersPerPolicy:      1,
+		DependentsPerPolicy: 0,
+	}
+
+	fixtures := models.CreatePolicyFixtures(as.DB, fixConfig)
+
+	for _, p := range fixtures.Policies {
+		assert.Nil(as.T(), p.LoadMembers(as.DB, false))
```
Schparky, line 26:
```suggestion
		as.NoError(p.LoadMembers(as.DB, false))
```

application/actions/policies_test.go
```diff
@@ -0,0 +1,98 @@
...
+		assert.Nil(as.T(), p.LoadDependents(as.DB, false))
```
Schparky, line 27:
```suggestion
		as.NoError(p.LoadDependents(as.DB, false))
```

application/actions/policies_test.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	// alias a couple users
+	appAdmin := fixtures.Policies[0].Members[0]
+	normalUser := fixtures.Policies[1].Members[0]
+
+	// change user 0 to an admin
+	appAdmin.AppRole = models.AppRoleAdmin
+	err := appAdmin.Update(as.DB)
+	assert.Nil(as.T(), err, "failed to make first policy user an app admin")
```
Schparky, line 37:
```suggestion
	as.NoError(err, "failed to make first policy user an app admin")
```

Schparky (approved):

PR #19, opened 2021-08-04
test fixtures

Schparky:


CONTRIBUTING.md
```diff
@@ -0,0 +1,129 @@
...
+function.
+
+Complex queries and joins can be accomplished using the model fields and
+iterating over the attached lists. This ends up being more complex and
+difficult to read. We have determined it is better to use raw SQL in these
+situations. For example:
+
+```go
+    var t Threads
+    query := DB.Q().LeftJoin("thread_participants tp", "threads.id = tp.thread_id")
```
Baggerone, line 83: Would it be good to update the object names here?

application/models/testutils.go
```diff
@@ -1,16 +1,40 @@
...
 	"github.com/gobuffalo/buffalo"
+	"github.com/gobuffalo/nulls"
 	"github.com/gobuffalo/pop/v5"
-
 	"github.com/silinternational/riskman-api/domain"
 )
 
+type FixturesConfig struct {
+	ItemsPerPolicy      int
+	Policies            int
```
Baggerone, line 22: Maybe something like `PolicyCount` (to distinguish it from a slice of Policy objects)?

application/models/testutils.go
```diff
@@ -36,42 +60,185 @@ func CreateTestContext(user User) buffalo.Context {
...
+		policies[i].CostCenter = randStr(10)
+		policies[i].HouseholdID = randStr(10)
+		MustCreate(tx, &policies[i])
+
+		f := CreatePolicyUserFixtures(tx, policies[i], config.UsersPerPolicy)
+		users = append(users, f.Users...)
+		policyUsers = append(policyUsers, f.PolicyUsers...)
+
+		pDependents := CreatePolicyDependentFixtures(tx, policies[i], config.DependentsPerPolicy).PolicyDependents
+		policyDependents = append(policyDependents, pDependents...)
```
Baggerone, line 164: Does this work, as opposed to something like pDFixtures.PolicyDependants?

Baggerone (approved):

CONTRIBUTING.md
Schparky: as an example, I think I'll save time and not look for a new example in this project...maybe later

application/models/testutils.go
Schparky: heh, it looks weird because of the long line. changing to make it easier to read

CONTRIBUTING.md
```diff
@@ -0,0 +1,129 @@
...
+
+[IDE configuration](#ide-configuration)
+
+[Coding Style](#coding-style)
+
+
+## IDE configuration
+
+### Custom tags
+Add "development" to the "Custom tags" setting in "Go - Build Tags & Vendoring" section of Goland Preferences. This ensures that the test fixture generation functions in `testutils.go` are not ignored by the IDE.
```
jason-jackson, line 13: Do we want to add VSCod(e|ium) to this?

`Add "Development" to the "Go:Build Tags" in VS Code extension preferences.`

PR #23, opened 2021-08-04
Stub out a listener for creating an initial policy for a new user

Baggerone:


application/listeners/listeners.go
```diff
@@ -0,0 +1,100 @@
...
+
+	if err := user.CreateInitialPolicy(nil); err != nil {
+		domain.ErrLogger.Printf("Failed to create initial policy in createUserPolicy, %s", err)
+		return
+	}
+
+}
+
+func getID(p events.Payload) (uuid.UUID, error) {
+	i, ok := p["id"]
```
Schparky, line 74: make a const for "id"?

application/models/user.go
```diff
@@ -166,3 +181,33 @@ func ConvertPolicyMembers(tx *pop.Connection, us Users) (api.PolicyMembers, erro
 
 	return members, nil
 }
+
+// CreateInitialPolicy is a hack to create an initial policy for a new user
+func (u *User) CreateInitialPolicy(tx *pop.Connection) error {
+	if u.ID == uuid.Nil {
```
Schparky, line 187:
```suggestion
	if u == nil || u.ID == uuid.Nil {
```

PR #24, opened 2021-08-05
Feature/update policy

fillup:


application/actions/policies_test.go
```diff
@@ -94,3 +94,103 @@ func (as *ActionSuite) Test_PoliciesList() {
...
+				as.Contains(body, tt.wantInBody)
+			}
+			if tt.notWantInBody != "" {
+				as.NotContains(body, tt.notWantInBody)
+			}
+
+			if res.Code != http.StatusOK {
+				return
+			}
+			fmt.Println(body)
```
Schparky, line 190: should this Println be removed now? And maybe add the body to one or both of the later test error messages?

Schparky (approved):

application/actions/policies_test.go
fillup: yes, thanks

application/actions/actions.go
```diff
@@ -35,9 +46,10 @@ func reportError(c buffalo.Context, err error) error {
 	appErr.LoadTranslatedMessage(c)
 
 	// clear out debugging info if not in development or test
-	if domain.Env.GoEnv != "development" && domain.Env.GoEnv != "test" {
+	if domain.Env.GoEnv == "development" || domain.Env.GoEnv == "test" {
+		appErr.DebugMsg = appErr.Err.Error()
+	} else {
 		appErr.Extras = map[string]interface{}{}
-		appErr.DebugMsg = ""
```
Baggerone, line 40: Is it safer to keep this?

application/actions/policies_test.go
```diff
@@ -94,3 +94,103 @@ func (as *ActionSuite) Test_PoliciesList() {
...
+			wantInBody: fixtures.Policies[1].ID.String(),
+		},
+		{
+			name:   "other person's policy",
+			actor:  normalUser,
+			policy: fixtures.Policies[0],
+			update: api.PolicyUpdate{
+				HouseholdID: "09876",
+			},
+			wantStatus:    http.StatusForbidden,
```
Baggerone, line 156: Should these be http.StatusNotFound? I thought that's the kind of thing we were trying to do in the other projects.

Baggerone (approved):

application/actions/policies.go
```diff
@@ -47,3 +51,47 @@ func policiesListMine(c buffalo.Context) error {
 
 	return c.Render(http.StatusOK, r.JSON(apiPolicies))
 }
+
+func policiesUpdate(c buffalo.Context) error {
+	tx := models.Tx(c)
+	policy := getReferencedPolicyFromCtx(c)
```
Baggerone, line 57: Should this check for nil?

fillup: we could, although the way authz middleware works, it shouldn't be possible to get into an action and it not be present

application/models/policy.go
```diff
@@ -25,10 +25,10 @@ var ValidPolicyTypes = map[api.PolicyType]struct{}{
 type Policy struct {
 	ID          uuid.UUID      `db:"id"`
 	Type        api.PolicyType `db:"type" validate:"policyType"`
-	HouseholdID string         `db:"household_id"`
-	CostCenter  string         `db:"cost_center"`
-	Account     string         `db:"account"`
-	EntityCode  string         `db:"entity_code"`
+	HouseholdID string         `db:"household_id" validate:"required_if=Type Household"`
```
fillup, line 28: @Baggerone @jason-jackson I just wanted to point out how this works. It took a bit to find in the docs, but this is how we make one field required if another field has a specific value. I believe we can chain additional validations as well.

application/actions/policies.go
Schparky: Anything is possible in development, haha. I encountered it while I was debugging that pointer vs non-pointer issue.

PR #26, opened 2021-08-05
list item categories

fillup:


Baggerone (approved):
Will you do authz later?

jason-jackson (approved):

fillup:
There is no AuthZ needed for that endpoint, later when we add abilities to add/update/etc. item categories we can add it.

PR #27, opened 2021-08-05
create dependent endpoint

Schparky:


application/actions/dependents.go
```diff
@@ -13,14 +13,33 @@ func dependentsList(c buffalo.Context) error {
...
+	policy.LoadDependents(tx, false)
 
+	return renderOk(c, models.ConvertPolicyDependents(tx, policy.Dependents))
+}
+
+func dependentsCreate(c buffalo.Context) error {
+	policy := getReferencedPolicyFromCtx(c)
+	if policy == nil {
+		err := errors.New("policy not found in route")
+		return reportError(c, api.NewAppError(err, api.ErrorPolicyNotFound, api.CategoryUser))
```
fillup, line 29: I think category of error in this case should be internal server error

fillup (approved):

PR #37, opened 2021-08-06
API endpoint to add an item

Baggerone:


Schparky (approved):

application/domain/domain.go
```diff
@@ -64,6 +64,10 @@ const (
 	TypeUser            = "users"
 )
 
+const (
+	DateFormat = "2006-01-02"
```
fillup, line 68: maybe the name for this constant should be more descriptive since all the dates we'll return will be RFC3339 or whatever. could this just be something like `ShortDateFormat`?

fillup (approved):

PR #38, opened 2021-08-06
list claims

Schparky:
(no unit test at this time)

### DIFF
--- a/application/actions/claims.go
+++ b/application/actions/claims.go
@@ -1,0 +1,30 @@
+package actions
+
+import (
+	"net/http"
+
+	"github.com/gobuffalo/buffalo"
+
+	"github.com/silinternational/riskman-api/domain"
+	"github.com/silinternational/riskman-api/models"
+)
+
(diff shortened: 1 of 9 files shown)


application/actions/claims.go
```diff
@@ -0,0 +1,30 @@
...
+import (
+	"net/http"
+
+	"github.com/gobuffalo/buffalo"
+
+	"github.com/silinternational/riskman-api/domain"
+	"github.com/silinternational/riskman-api/models"
+)
+
+func claimsList(c buffalo.Context) error {
```
fillup, line 12: We need some extra logic in here to check the calling user and scoping the list of claims appropriately. Same as list policies.

fillup (changes requested):

Baggerone (approved):

application/actions/claims.go
Schparky: For now it does not allow a user to list claims. Will come back to it after doing the other claims endpoints

PR #44, opened 2021-08-10
Item update

Baggerone:


application/models/item.go
```diff
@@ -57,6 +57,10 @@ func (i *Item) Create(tx *pop.Connection) error {
 	return create(tx, i)
 }
 
+func (i *Item) Update(tx *pop.Connection) error {
```
jason-jackson, line 60: Should this also log to policy_history table too?

Schparky (approved):

application/models/item.go
Baggerone: Maybe eventually. I don't think we've started those yet.

PR #46, opened 2021-08-10
Remove item

Baggerone:


application/models/item.go
```diff
@@ -69,6 +70,24 @@ func (i *Item) FindByID(tx *pop.Connection, id uuid.UUID) error {
 	return tx.Find(i, id)
 }
 
+func (i *Item) SafeDelete(tx *pop.Connection, actor User) error {
+	if !i.IsActorAllowedTo(tx, actor, PermissionDelete, "", nil) {
```
Schparky, line 74: Isn't this already checked by the middleware?

application/models/item.go
```diff
@@ -69,6 +70,24 @@ func (i *Item) FindByID(tx *pop.Connection, id uuid.UUID) error {
...
+	clICount, err := tx.Where("item_id = ?", i.ID).Count(&clItems)
+	if err != nil {
+		return err
+	}
+
+	if clICount > 0 {
+		return errors.New("deleting the item is not allowed, since it has an associated claim item.")
+	}
+
+	return tx.Destroy(i)
```
fillup, line 88: I'm not sure if we'll ever delete an item, at least not once it has had a billable event. Perhaps only on the day it is created or before an entry is added into whatever table we track premium charges in. Rather than deleting, the coverage status should just change to `Inactive` (which I don't believe exists in the code yet), and then we'll need to calculate partial refunds and such.

application/models/item.go
```diff
@@ -69,6 +70,24 @@ func (i *Item) FindByID(tx *pop.Connection, id uuid.UUID) error {
...
+	}
+
+	clItems := ClaimItems{}
+	clICount, err := tx.Where("item_id = ?", i.ID).Count(&clItems)
+	if err != nil {
+		return err
+	}
+
+	if clICount > 0 {
+		return errors.New("deleting the item is not allowed, since it has an associated claim item.")
```
Schparky, line 85: Maybe this should be an `AppError` with type "User"

Schparky (approved):

application/actions/items.go
```diff
@@ -155,6 +155,23 @@ func itemsUpdate(c buffalo.Context) error {
 	return c.Render(http.StatusOK, r.JSON(output))
 }
 
+func itemsRemove(c buffalo.Context) error {
```
Schparky, line 158: add swagger?

PR #49, opened 2021-08-11
swagger docs

Schparky:


application/actions/claims.go
```diff
@@ -23,6 +37,23 @@ func claimsList(c buffalo.Context) error {
...
 
+// swagger:operation GET /claims/{id} Claims ClaimsView
+//
+// ClaimsView
+//
+// view a specific claim
+//
+// ---
+// parameters:
+// - name: id
```
Baggerone, line 48: At one point, we thought we didn't need to have a note about these obvious path parameters. 
Do you prefer having them?

Schparky: I don't recall, but it seemed to be necessary for the generated docs

Schparky: Also, since they're all `id` it seems to add a bit of clarity

application/actions/items.go
```diff
@@ -56,6 +99,29 @@ func itemsCreate(c buffalo.Context) error {
 	return c.Render(http.StatusOK, r.JSON(output))
 }
 
+// swagger:operation PUT /policies/{id}/items PolicyItems PolicyItemsUpdate
```
Baggerone, line 102:
```suggestion
// swagger:operation PUT /items/{id} PolicyItems PolicyItemsUpdate
```

application/actions/claims.go
fillup: is there any annotation for these to indicate that it should be a uuid v4?

Baggerone (approved):

application/actions/items.go
fillup, line 102: This should probably be `PUT /items/{id}` if we already have the item ID, and the update model might need to be different than a create. We need to think through what we'd allow to be edited over time as some properties may need to trigger new workflows, like changing the item value to something above an auto-approval amount would need to go through approval, etc.

application/api/items.go
```diff
@@ -21,49 +26,103 @@ type Items []Item
...
+	// model
+	Model string `json:"model"`
+
+	// serial number
+	SerialNumber string `json:"serial_number"`
+
+	// coverage amount (USD)
+	CoverageAmount int `json:"coverage_amount"`
 
+	// date (yyyy-mm-dd) of item's purchase
```
fillup, line 67: any idea if this short date format same as goswagger's `swagger:strfmt date` format? the docs don't provide an example that I could see. https://goswagger.io/use/spec/strfmt.html

fillup (approved):

application/api/items.go
Schparky: image

application/actions/items.go
Schparky: thanks

application/actions/claims.go
Schparky: nope

PR #50, opened 2021-08-11
my claims

Schparky:


fillup (approved):

application/actions/claims.go
```diff
@@ -26,17 +25,30 @@ import (
...
+
+	if user.IsAdmin() {
+		return claimsListAll(c)
+	}
+
+	return claimsListMine(c)
+}
 
-	// TODO: list only current user's claims
+func claimsListAll(c buffalo.Context) error {
```
Baggerone, line 37: I'm guessing this will be unreasonably large once we include historical data.

Schparky: Possibly. Will likely need to do some sort of paging

PR #51, opened 2021-08-11
Fix logout error and add swagger for authn

Baggerone:


application/actions/auth.go
```diff
@@ -51,6 +51,19 @@ var samlConfig = saml.Config{
 	AttributeMap:                nil,
 }
 
+// swagger:operation POST /auth/login Authentication AuthLogin
+//
+// Start the SAML login process
```
Schparky, line 56: If you provide a description, you should also provide a title. So like this:
```
// swagger:operation POST /auth/login Authentication AuthLogin
//
// AuthLogin
//
// Start the SAML login process
//
// ---
```
Otherwise, goswagger uses the text between `swagger:operation` and `---` as a title.

application/actions/auth.go
```diff
@@ -214,14 +227,25 @@ func getLoginSuccessRedirectURL(authUser auth.User, returnTo string) string {
 	return uiURL + returnTo + params
 }
 
-// authDestroy uses the bearer token to find the user's access token and
-//  calls the appropriate provider's logout function.
+// swagger:operation GET /auth/logout Authentication AuthLogout
+//
+// Logout of application
```
Schparky, line 232: Same as with /auth/login.

application/actions/auth.go
```diff
@@ -214,14 +227,25 @@ func getLoginSuccessRedirectURL(authUser auth.User, returnTo string) string {
...
 	tokenParam := c.Param(LogoutToken)
 	if tokenParam == "" {
 		return reportErrorAndClearSession(c, &api.AppError{
-			HttpStatus: http.StatusInternalServerError,
-			Key:        api.ErrorCreatingAccessToken,
+			HttpStatus: http.StatusBadRequest,
+			Key:        api.ErrorMissingLogoutToken,
 			Message:    LogoutToken + " is required to logout",
 		})
 	}
```
Schparky, line 251: While you're at it, should /auth/callback have swagger docs?

Schparky (approved):

application/actions/auth.go
Baggerone: I don't think it's needed, since that's just a part of the automated saml handshake process.

PR #52, opened 2021-08-12
Start converting errors to appErrors in models

Baggerone:
Is this the way we want to do it?


application/models/useraccesstoken.go
```diff
@@ -62,11 +64,20 @@ func (u *UserAccessToken) ValidateUpdate(tx *pop.Connection) (*validate.Errors,
...
-		return err
+func (u *UserAccessToken) DeleteByBearerToken(tx *pop.Connection, bearerToken string) *api.AppError {
+	if appErr := u.FindByBearerToken(tx, bearerToken); appErr != nil {
+		return appErr
+	}
+	if err := tx.Destroy(u); err != nil {
+		appErr := api.AppError{
+			Err:      err,
+			Key:      api.ErrorDeletingAccessToken,
+			Category: api.CategoryDatabase,
```
Schparky, line 75: Just wondering if we still want a separate `CategoryDatabase` since it generates a 500 just like `CategoryInternal`. In which case, it could just `panic`, correct?

application/models/useraccesstoken.go
```diff
@@ -62,11 +64,20 @@ func (u *UserAccessToken) ValidateUpdate(tx *pop.Connection) (*validate.Errors,
 }
 
 // DeleteByBearerToken uses a sha256.Sum256 of the bearerToken to find which UserAccessToken to delete
-func (u *UserAccessToken) DeleteByBearerToken(tx *pop.Connection, bearerToken string) error {
-	if err := u.FindByBearerToken(tx, bearerToken); err != nil {
-		return err
+func (u *UserAccessToken) DeleteByBearerToken(tx *pop.Connection, bearerToken string) *api.AppError {
```
Schparky, line 67: I think this should return `error` (not `*api.AppError`)

application/models/useraccesstoken.go
```diff
@@ -83,13 +94,23 @@ func (u *UserAccessToken) DeleteIfExpired(tx *pop.Connection) (bool, error) {
 }
 
 // FindByBearerToken uses a sha256.Sum256 of the bearerToken to find the corresponding UserAccessToken
-func (u *UserAccessToken) FindByBearerToken(tx *pop.Connection, bearerToken string) error {
+func (u *UserAccessToken) FindByBearerToken(tx *pop.Connection, bearerToken string) *api.AppError {
```
Schparky, line 97:
```suggestion
func (u *UserAccessToken) FindByBearerToken(tx *pop.Connection, bearerToken string) error {
```

application/models/useraccesstoken.go
```diff
@@ -83,13 +94,23 @@ func (u *UserAccessToken) DeleteIfExpired(tx *pop.Connection) (bool, error) {
...
-		return fmt.Errorf("failed to find access token '%s...', %s", bearerToken[0:l], err)
+
+		appErr := api.AppError{
+			Err:      err,
+			Key:      api.ErrorFindingAccessToken,
+			Category: api.CategoryUser,
+			Message:  fmt.Sprintf("failed to find access token '%s...'", bearerToken[0:l]),
+		}
+		if domain.IsOtherThanNoRows(err) {
+			appErr.Category = api.CategoryDatabase
```
Schparky, line 111: Same comment about `CategoryDatabase` vs `CategoryInternal` vs `panic`

Schparky (commented):
On the right track. Just added a few comments for discussion

application/models/useraccesstoken.go
Baggerone: I was wondering about that, but I was thinking that it would be good to make it explicit, so that when we call it, we know we don't need to convert the result to an AppError.

What's the benefit of having just an `error` returned?

Schparky: Good point. I am not really sure, but it seems more "standard" to use `error`. I probably need to do more reading on the subject.

Schparky (approved):

jason-jackson:
I don't have strong feelings either way. Given this is my first project with Buffalo, I'll just go with the flow.  It doesn't look bad to me.

application/models/useraccesstoken.go
Schparky: I found this in the Go FAQ: https://golang.org/doc/faq#nil_error. The second-to-last paragraph contains the statement "It's a good idea for functions that return errors always to use the error type in their signature".

Schparky (approved):

PR #57, opened 2021-08-17
Add new statuses for item coverage and item claim and some validation

Baggerone:
Does this look helpful?

### DIFF
--- a/application/actions/items_test.go
+++ b/application/actions/items_test.go
@@ -290,7 +290,7 @@ func (as *ActionSuite) Test_ItemsUpdate() {
 		SerialNumber:      "MM1234",
 		CoverageAmount:    oldItem.CoverageAmount,
 		PurchaseDate:      "2006-01-02",
-		CoverageStatus:    api.ItemCoverageStatusDraft,
+		CoverageStatus:    api.ItemCoverageStatusInactive,
 		CoverageStartDate: "2006-01-03",
 	}
 
--- a/application/models/item.go
+++ b/application/models/item.go
@@ -1,6 +1,8 @@
 package models
 
 import (
+	"errors"
+	"fmt"
 	"net/http"
 	"time"
 
@@ -17,6 +19,7 @@ import (
 var ValidItemCoverageStatuses = map[api.ItemCoverageStatus]struct{}{
 	api.ItemCoverageStatusDraft:    {},
 	api.ItemCoverageStatusPending:  {},
+	api.ItemCoverageStatusRevision: {},
 	api.ItemCoverageStatusApproved: {},
 	api.ItemCoverageStatusDenied:   {},
 	api.ItemCoverageStatusInactive: {},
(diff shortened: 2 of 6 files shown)


application/actions/items_test.go
```diff
@@ -290,7 +290,7 @@ func (as *ActionSuite) Test_ItemsUpdate() {
 		SerialNumber:      "MM1234",
 		CoverageAmount:    oldItem.CoverageAmount,
 		PurchaseDate:      "2006-01-02",
-		CoverageStatus:    api.ItemCoverageStatusDraft,
+		CoverageStatus:    api.ItemCoverageStatusInactive,
```
jason-jackson, line 293: Was this meant to change? If so, why is it a "good item"

application/models/item.go
```diff
@@ -136,6 +150,50 @@ func (i *Item) IsActorAllowedTo(tx *pop.Connection, actor User, perm Permission,
 	return false
 }
 
+func itemStatusTransitions() map[api.ItemCoverageStatus][]api.ItemCoverageStatus {
```
jason-jackson, line 153: Would a map of maps be better?

application/models/item.go
```diff
@@ -113,8 +126,9 @@ func (i *Item) isNewEnough() bool {
 // Inactivate sets the item's CoverageStatus to Inactive
 //  TODO deal with coverage payment changes
 func (i *Item) Inactivate(tx *pop.Connection) error {
+	status1 := i.CoverageStatus
```
jason-jackson, line 129: Personally, I like `oldStatus` or `currentStatus` or even `status` better

jason-jackson (approved):

application/models/item.go
```diff
@@ -58,7 +61,17 @@ func (i *Item) Create(tx *pop.Connection) error {
 	return create(tx, i)
 }
 
-func (i *Item) Update(tx *pop.Connection) error {
+func (i *Item) Update(tx *pop.Connection, oldStatus api.ItemCoverageStatus) error {
```
Schparky, line 64: I think this works. I wonder, though, if it could/should take the new status as the parameter rather than have the caller set the new status on the object.

Schparky (approved):

application/actions/items_test.go
Baggerone: It had to be an acceptable status transition.  It's still a "good item", because it doesn't have any errors.

application/models/item.go
Baggerone: I think I would use a map if there were lots of elements in each slice. But with so few, I doubt it's worth it.

application/models/item.go
Baggerone: :+1:

application/models/item.go
Baggerone: I don't know if it's worth it. Right now in the action it's called after convertItemApiInput() is called which already sets the status to the new one.

We would need to make that stop setting the status.

Schparky: OK, that's fine.

PR #60, opened 2021-08-17
Add two fields to user EmailOverride and Location

Baggerone:


application/migrations/20210714203100_create_users.up.fizz
```diff
@@ -1,10 +1,12 @@
 create_table("users") {
 	t.Column("id", "uuid", {primary: true})
 	t.Column("email", "string", {})
+	t.Column("email_override", "string", {})
```
Schparky, line 4:
```suggestion
	t.Column("email_override", "string", {"default": ""})
```

application/migrations/20210714203100_create_users.up.fizz
```diff
@@ -1,10 +1,12 @@
...
 	t.Column("first_name", "string", {})
 	t.Column("last_name", "string", {})
 	t.Column("is_blocked", "bool", {})
 	t.Column("last_login_utc", "timestamp", {})
+	t.Column("location", "string", {})
```
Schparky, line 9:
```suggestion
	t.Column("location", "string", {"default": ""})
```

Schparky (approved):
Should `Location` be required at the database level?

application/migrations/20210714203100_create_users.up.fizz
Baggerone: :+1:

application/migrations/20210714203100_create_users.up.fizz
Baggerone: :+1:

jason-jackson (approved):
Generally I think new migration is more kosher. Also, do we have a way to tell if EmailOverride is present, but not selected (for example, if they fill it in but change back to old email), or do we expect UI handle that? That is, should APIs have another field for selection?

PR #63, opened 2021-08-18
Include claim items in claims API responses

fillup:


Baggerone (approved):

application/models/testutils.go
```diff
@@ -121,6 +122,20 @@ func createClaimFixture(tx *pop.Connection, policyID uuid.UUID) Claim {
...
 	}
 	MustCreate(tx, &claim)
+
+	icFixtures := CreateCategoryFixtures(tx, claimItemCount)
+
+	claim.ClaimItems = make(ClaimItems, claimItemCount)
+	for i := range claim.ClaimItems {
+		item := createItemFixture(tx, policyID, icFixtures.ItemCategories[i].ID)
+		claim.ClaimItems[i] = ClaimItem{
+			ClaimID: claim.ID,
```
Schparky, line 132: Not sure it matters a lot, but I've been adding data on all of the fields. Just give me a bit of comfort thinking it's closer to "real".

application/models/testutils.go
```diff
@@ -288,6 +303,8 @@ func DestroyAll() {
 	destroyTable(&users)
 
 	// delete all Claims and ClaimItems
+	var claimItems ClaimItems
```
Schparky, line 306: The cascading delete is supposed to do this. Did you find otherwise?

application/models/testutils.go
```diff
@@ -82,7 +83,7 @@ func CreateItemFixtures(tx *pop.Connection, config FixturesConfig) Fixtures {
 
 		for k := 0; k < config.ClaimsPerPolicy; k++ {
 			idx := i*config.ClaimsPerPolicy + k
-			claims[idx] = createClaimFixture(tx, policies[i].ID)
+			claims[idx] = createClaimFixture(tx, policies[i].ID, config.ClaimItemsPerClaim)
```
Schparky, line 86: My plan was to pass the same config into all fixture generators. Since only one value is needed, this might be better though.

Schparky (approved):

application/models/testutils.go
fillup: it did not, I just got a ton of foreign key constraint errors

application/models/testutils.go
fillup: that'd be fine/good,  it just wasn't doing that currently so I just sent in what I needed. but I'd be happy to change it

application/models/testutils.go
fillup: sure, can do

PR #64, opened 2021-08-18
list claim event types

Schparky:


application/actions/app.go
```diff
@@ -137,6 +138,17 @@ func App() *buffalo.App {
 		claimsGroup.GET(idRegex, claimsView)
 		claimsGroup.PUT(idRegex, claimsUpdate)
 
+		// config
+		configGroup := app.Group("/config")
```
jason-jackson, line 142: Should this be added to domain? `domain.TypeConfig`

jason-jackson (approved):

Baggerone (approved):

application/actions/app.go
```diff
@@ -137,6 +138,17 @@ func App() *buffalo.App {
...
+		configGroup.Middleware.Skip(AuthN, claimEventTypes)
```
fillup, line 143: Do we really want/need claim event types to be publicly accessible? Is there harm in keeping AuthN in place? The data from this endpoint isn't used or needed until the user is authenticated and creating a claim. I agree it doesn't require AuthZ though.

Schparky: I just thought it was not needed since it's not sensitive data, but I'm fine with leaving AuthN on it.

application/actions/app.go
Schparky: Doesn't need to be, those types are for AuthZ. Also, it's not really a "type". If you think it should be a `const` I don't mind adding it, though.

application/actions/app.go
fillup: ok sure, lets leave AuthN on for now, and if we have a config endpoint that needs to be loaded pre-login, then we can skip authn then

fillup, line 143:
```suggestion
```

PR #68, opened 2021-08-20
add reference_number to claims

fillup:
also updated test utils to use a models Create method since that is where the reference number is generated

### DIFF
--- a/application/domain/domain.go
+++ b/application/domain/domain.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"io"
 	"log"
+	"math/rand"
 	"net/http"
 	"os"
 	"regexp"
@@ -320,4 +321,14 @@ func IsStringInSlice(needle string, haystack []string) bool {
 	}
 
 	return false
+}
+
+var letters = []rune("ABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789")
+
+func RandomString(n int) string {
+	b := make([]rune, n)
+	for i := range b {
+		b[i] = letters[rand.Intn(len(letters))]
+	}
+	return string(b)
 }
--- a/application/models/claim.go
+++ b/application/models/claim.go
@@ -15,6 +15,10 @@ import (
 	"github.com/silinternational/riskman-api/domain"
 )
 
+const (
+	ClaimReferenceNumberLength = 6
+)
+
 var ValidClaimEventTypes = map[api.ClaimEventType]struct{}{
 	api.ClaimEventTypeTheft:           {},
 	api.ClaimEventTypeImpact:          {},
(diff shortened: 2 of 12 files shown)


application/domain/domain.go
```diff
@@ -321,3 +322,13 @@ func IsStringInSlice(needle string, haystack []string) bool {
 
 	return false
 }
+
+var letters = []rune("ABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789")
```
Schparky, line 326: Any reason to have this at package scope? If not, move it inside `RandomString`.

Schparky (approved):

application/models/claim.go
```diff
@@ -60,6 +65,7 @@ func (c *Claim) Validate(tx *pop.Connection) (*validate.Errors, error) {
 
 // Create stores the Claim data as a new record in the database.
 func (c *Claim) Create(tx *pop.Connection) error {
+	c.ReferenceNumber = domain.RandomString(ClaimReferenceNumberLength)
```
jason-jackson, line 68: Should more be done to ensure a given Claim reference number is not repeated? As this would fail if there's any duplication, unless I'm missing something.

fillup: There is a unique clause on the column in the db which would prevent a duplicate from being created, but I suppose it would not be graceful or have any way to recover, so perhaps that's worth some thought.

application/domain/domain.go
```diff
@@ -321,3 +322,12 @@ func IsStringInSlice(needle string, haystack []string) bool {
 
 	return false
 }
+
+func RandomString(n int) string {
```
jason-jackson, line 326: Should we try to avoid bad words? If so, it might be easier to use C<number of length X> instead of this.

application/models/claim.go
Baggerone: Maybe have a loop that gets a new ref # and checks whether it exists. If it doesn't exist, exit the loop.

Baggerone (approved):

application/models/claim.go
```diff
@@ -247,3 +251,22 @@ func ConvertClaimCreateInput(input api.ClaimCreateInput) Claim {
...
+
+func uniqueClaimReferenceNumber(tx *pop.Connection) string {
+	attempts := 0
+	for {
+		// create reference number in format CAB1234
+		ref := fmt.Sprintf("C%s%s",
+			domain.RandomString(2, "ABCDEFGHIJKLMNOPQRSTUVWXYZ"),
+			domain.RandomString(4, "1234567890"))
+		var claim Claim
+		err := claim.FindByReferenceNumber(tx, ref)
```
Baggerone, line 263: This may not be worth changing, but doing a `count` query may be a little cheaper and I assume it would only have one kind of error.

The way you have it, if there is a real db error (other than no rows), it will keep trying 100 times, right?

PR #70, opened 2021-08-23
Feature/item submit

Baggerone:


application/actions/items_test.go
```diff
@@ -249,6 +249,107 @@ func (as *ActionSuite) Test_ItemsAdd() {
...
+				`"purchase_date":"` + revisionItem.PurchaseDate.Format(domain.DateFormat) + `"`,
+				`"coverage_status":"` + string(api.ItemCoverageStatusPending),
+				`"category":{"id":"` + iCatID.String(),
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		as.T().Run(tt.name, func(t *testing.T) {
+			req := as.JSON("/items/%s/submit", tt.oldItem.ID.String())
```
jason-jackson, line 330: Since we're importing domain, using domain.ItemType in the first part of the req would increase portability.

jason-jackson (approved):

application/actions/items.go
```diff
@@ -152,6 +152,38 @@ func itemsUpdate(c buffalo.Context) error {
 	return c.Render(http.StatusOK, r.JSON(output))
 }
 
+// swagger:operation POST /items/{id} PolicyItems PolicyItemsSubmit
```
fillup, line 155:
```suggestion
// swagger:operation POST /items/{id}/submit PolicyItems PolicyItemsSubmit
```

application/actions/items_test.go
fillup: that would at least help in multiple dependencies/tests on those Type values staying consistent.

Schparky (approved):

PR #72, opened 2021-08-23
data import

Schparky:
Though not final, I think it's close enough for review. In hindsight, I probably should have submitted it sooner.

Note that the new migration drops the unique key on `household_id` so the import doesn't fail. Before merging, I'll remove that.


fillup (approved):

application/grifts/types.go
```diff
@@ -0,0 +1,162 @@
...
+	Policies       []LegacyPolicy       `json:"policies"`
+	PolicyTypes    []PolicyType         `json:"PolicyType"`
+	Maintenance    []Maintenance        `json:"Maintenance"`
+	JournalEntries []JournalEntry       `json:"tblJEntry"`
+	ItemCategories []LegacyItemCategory `json:"item_categories"`
+	RiskCategories []LegacyRiskCategory `json:"risk_categories"`
+	LossReasons    []LossReason         `json:"LossReason"`
+}
+
+type LegacyUsers struct {
```
Baggerone, line 16: Just observing that this is plural when the others are singular.

application/models/validation.go
```diff
@@ -115,3 +115,35 @@ func validateItemCoverageStatus(field validator.FieldLevel) bool {
 	}
 	return false
 }
+
+func claimStructLevelValidation(sl validator.StructLevel) {
+	claim := sl.Current().Interface().(Claim)
```
Baggerone, line 120: Don't you want to check that it's really a Claim (here and in next function for ClaimItem)?

Baggerone (approved):

application/models/validation.go
Schparky: I'm not concerned about it because that would only happen if this validation was registered to the wrong type (or some other bug). So a panic is appropriate. If you prefer, I don't mind adding that here.

application/grifts/types.go
Schparky: Thanks, I'll change it

application/models/validation.go
Baggerone: I'm thinking a non-descriptive panic might not be too helpful for troubleshooting.

PR #91, opened 2021-08-30
Admin requests revision on a claim

Baggerone:


application/actions/claims_test.go
```diff
@@ -28,14 +28,9 @@ func (as *ActionSuite) Test_ClaimsList() {
 	fixtures := models.CreateItemFixtures(as.DB, fixConfig)
 
 	// alias a couple users
-	appAdmin := fixtures.Policies[0].Members[0]
+	appAdmin := models.CreateAdminUser(as.DB)
```
Schparky, line 31: technically, the comment "alias a couple users" doesn't apply to the`appAdmin` line any longer

application/models/claim.go
```diff
@@ -252,6 +259,32 @@ func (c *Claim) SubmitForApproval(tx *pop.Connection) error {
...
 
+// RequestRevision changes the status of the claim to Revision
+//   provided that the current status is Review1 or Review3.
+// TODO record the particular revisions that are needed
+// TODO emit an appropriate event
+func (c *Claim) RequestRevision(tx *pop.Connection) error {
+	oldStatus := c.Status
+
+	c.LoadClaimItems(tx, false)
+	if len(c.ClaimItems) == 0 {
```
Schparky, line 270: I see this check being repeated. Would it be possible to put it in the `Update` function? Struct-level validation might be better, but there's no context there to get the database transaction.

Schparky (approved):

application/models/claim.go
Baggerone: :+1: 
I'm having it ignore the items if the status is still just Draft

PR #93, opened 2021-08-30
use real item categories in seed data

Schparky:
When the category data changes, after running the import tool, the item_categories SQL can be pasted into db.go from an adminer SQL export.

### DIFF
--- a/application/grifts/db.go
+++ b/application/grifts/db.go
@@ -2,6 +2,7 @@ package grifts
 
 import (
 	"fmt"
+	"regexp"
 	"time"
 
 	"github.com/gobuffalo/nulls"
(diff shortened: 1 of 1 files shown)


application/grifts/db.go
```diff
@@ -190,45 +191,46 @@ func createPolicyFixtures(fixUsers []*models.User) ([]*models.Policy, error) {
...
 }
 
-func createCategories() ([]*models.ItemCategory, error) {
-	riskCats := models.RiskCategories{}
-	err := models.DB.All(&riskCats)
-	if err != nil {
-		return []*models.ItemCategory{}, err
+func createCategories() ([]uuid.UUID, error) {
+	const itemCategoriesSql = `
+INSERT INTO "item_categories" ("id", "risk_category_id", "name", "help_text", "status", "auto_approve_max", "created_at", "updated_at", "legacy_id") VALUES
```
jason-jackson, line 196: What's the security implications of having this here?

Baggerone (approved):

application/grifts/db.go
Schparky: No security implications that I can imagine at this time.

PR #96, opened 2021-08-30
refactor convert functions

Schparky:
I wanted to make this change before making unit tests for these functions. I'm open to suggestions on the new name.

### DIFF
--- a/application/models/item.go
+++ b/application/models/item.go
@@ -323,33 +323,33 @@ func (i *Item) Load(tx *pop.Connection) {
 	}
 }
 
-func ConvertItem(tx *pop.Connection, item Item) api.Item {
-	item.Load(tx)
+func (i *Item) ConvertToAPI(tx *pop.Connection) api.Item {
+	i.Load(tx)
 	return api.Item{
-		ID:                item.ID,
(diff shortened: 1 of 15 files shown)


application/models/item.go
```diff
@@ -323,33 +323,33 @@ func (i *Item) Load(tx *pop.Connection) {
...
+		CreatedAt:         i.CreatedAt,
+		UpdatedAt:         i.UpdatedAt,
 	}
 }
 
-func ConvertItems(tx *pop.Connection, items Items) api.Items {
-	apiItems := make(api.Items, len(items))
-	for i, p := range items {
-		apiItems[i] = ConvertItem(tx, p)
+func (i *Items) ConvertToAPI(tx *pop.Connection) api.Items {
```
jason-jackson, line 349: Do you need to use `*Items`? Earlier for claims only non-pointers were used (see `Claims` and `ClaimItems`), while singular item used a pointer.

Note: I'm fine with either way, so if you'd rather change `Claims` and `ClaimItems` than everything below, that is ok too.

jason-jackson (approved):

Baggerone (approved):

PR #98, opened 2021-08-31
Emit claim events

Baggerone:
I left a little duplication in app.go for preapprove and request receipt. Is that OK?

### DIFF
--- a/application/listeners/listeners_test.go
+++ b/application/listeners/listeners_test.go
@@ -49,7 +49,7 @@ func (ts *TestSuite) Test_findObject() {
 	}{
 		{
 			name:    "find user",
-			payload: events.Payload{domain.EventPayloadID: user.ID.String()},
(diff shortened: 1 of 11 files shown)


application/listeners/listeners_test.go
```diff
@@ -64,6 +65,15 @@ func (ts *TestSuite) Test_findObject() {
 				"Name:" + item.Name,
 			},
 		},
+		{
+			name:    "find claim",
+			payload: events.Payload{domain.EventPayloadID: claim.ID.String()},
```
Schparky, line 70: I should have noticed this on the previous PR, but I think it would be fine (maybe better) to pass the ID as a `uuid.UUID` rather than `string`.

Schparky (approved):

application/listeners/listeners_test.go
Baggerone: :+1:

PR #103, opened 2021-09-01
make renderOk do the conversion to the api type

Schparky:


Baggerone:
I'm not so keen on this change.  I'm thinking we're not gaining much by it, and having methods returning interfaces is a drawback for me.

Schparky:
> I'm not so keen on this change. I'm thinking we're not gaining much by it, and having methods returning interfaces is a drawback for me.

OK, I'll close it then. No problem.

PR #105, opened 2021-09-02
Send different emails for submitted item based on whether it was auto approved

Baggerone:
Also, figured out  a way to test the emails that the listeners themselves send out.


fillup (approved):

application/domain/domain.go
```diff
@@ -111,7 +111,9 @@ const (
 
 //  Notification templates
 const (
-	MessageTemplateItemSubmitted = "item_submitted"
+	MessageTemplateItemSubmitted      = "item_submitted"
```
Schparky, line 114: Is the convention to have "Steward" on the end of messages that go to the Steward? If not, maybe it would be helpful to quickly identify which are user emails and which are admin emails.

application/listeners/listeners.go
```diff
@@ -13,6 +13,8 @@ import (
 	"github.com/silinternational/cover-api/models"
 )
 
+const NotifierKey = "notifier"
```
Schparky, line 16:
```suggestion
const EventPayloadNotifier = "notifier"
```
for consistency with `EventPayloadID`

application/listeners/item.go
```diff
@@ -23,27 +46,48 @@ func itemSubmitted(e events.Event) {
...
-		Data: map[string]interface{}{
-			"appName":        domain.Env.AppName,
-			"uiURL":          domain.Env.UIURL,
-			"itemMemberName": memberName,
-			"itemURL":        fmt.Sprintf("%s/items/%s", domain.Env.UIURL, item.ID),
-			"itemName":       item.Name,
-		},
+	notifiers := getNotifiersFromEventPayload(e.Payload)
+
+	if _, isApproved := e.Payload[string(api.ItemCoverageStatusApproved)]; isApproved {
```
Schparky, line 54: Would it work to just use `item.CoverageStatus === api.ItemCoverageStatusApproved` and not add a separate entry into the payload?

application/notifications/email_test.go
```diff
@@ -24,13 +24,13 @@ func (ts *TestSuite) TestSend() {
 		},
 	}
 	var emailService EmailService
-	var testService DummyEmailService
+	testService := DummyEmailService{}
```
Schparky, line 27: I'm curious about this change.

application/notifications/messages.go
```diff
@@ -11,3 +16,24 @@ type Message struct {
...
+		Data: map[string]interface{}{
+			"appName": domain.Env.AppName,
+			"uiURL":   domain.Env.UIURL,
+		},
+	}
+	return msg
+}
+
+// sets the msg ToName and ToEmail based on the steward's information
+func (m *Message) AddToSteward() {
```
Schparky, line 33:
```suggestion
func (m *Message) ToSteward() Message {
```
Then you can do this:
```
msg := notifications.NewEmailMessage().ToSteward()

File: application/notifications/messages.go [hunk 6]
```diff
@@ -11,3 +16,24 @@ type Message struct {
...
+	var steward models.User
+	steward.FindSteward(models.DB)
+
+	m.ToName = steward.Name()
+	m.ToEmail = steward.EmailOfChoice()
+}
```
Schparky, line 39:
```suggestion
	return *m
}
```

application/notifications/messages.go
```diff
@@ -11,3 +16,24 @@ type Message struct {
 	ToPhone   string
 	Subject   string
 }
+
+// NewEmailMessage returns a message with the FromEmail, the Data.appName and Data.uiURL already set
```
Schparky, line 20: These helpers are a good idea.

Schparky (approved):

application/listeners/item.go
Baggerone: Good point.

application/listeners/listeners.go
Baggerone: :+1:

application/domain/domain.go
Baggerone: :+1:

application/notifications/email_test.go
Baggerone: That was just a change I made when I was experimenting with the interface stuff and I forgot to revert it.  I'll revert it now.

application/notifications/messages.go
Baggerone: :+1:

PR #107, opened 2021-09-02
new endpoint: ClaimItemsUpdate

Schparky:


application/models/claimitem.go
```diff
@@ -60,24 +59,30 @@ func (c *ClaimItem) Validate(tx *pop.Connection) (*validate.Errors, error) {
...
+		err := fmt.Errorf("invalid claim item status transition from %s to %s", oldStatus, newStatus)
 		appErr := api.NewAppError(err, api.ErrorValidation, api.CategoryUser)
 		return appErr
 	}
-	return update(tx, c)
+
+	c.Status = newStatus
+	if newStatus == api.ClaimItemStatusDenied || newStatus == api.ClaimItemStatusRevision || newStatus == api.ClaimItemStatusApproved {
+		c.ReviewerID = nulls.NewUUID(user.ID)
+		c.ReviewDate = nulls.NewTime(time.Now())
```
jason-jackson, line 79: Should this be `.UTC()`?

application/models/claimitem.go
```diff
@@ -60,24 +59,30 @@ func (c *ClaimItem) Validate(tx *pop.Connection) (*validate.Errors, error) {
...
+	}
+	if err := update(tx, c); err != nil {
```
jason-jackson, line 81: Should ReviewerID and ReviewDate be rolled back as well?

application/models/claimitem.go
```diff
@@ -60,24 +59,30 @@ func (c *ClaimItem) Validate(tx *pop.Connection) (*validate.Errors, error) {
...
+		c.Status = oldStatus
+		return err
+	}
+	return nil
```
jason-jackson, line 85: Could change this to `return err` and remove return from above if statement.

jason-jackson (approved):

application/actions/claimitems_test.go
```diff
@@ -0,0 +1,111 @@
...
+			claimItem:  claimItem,
+			input:      input,
+			wantStatus: http.StatusOK,
+			wantInBody: claim.ID.String(),
+		},
+		{
+			name:       "admin user",
+			actor:      appAdmin,
+			claimItem:  claimItem,
+			input:      input,
```
Baggerone, line 76: It looks like this doesn't actually test the "new" values, since they will be the same as the previous test's.
Not that big a deal, though.

application/models/claimitem.go
```diff
@@ -60,24 +59,30 @@ func (c *ClaimItem) Validate(tx *pop.Connection) (*validate.Errors, error) {
...
-func (c *ClaimItem) Update(tx *pop.Connection, oldStatus api.ClaimItemStatus) error {
-	validTrans, err := isClaimItemTransitionValid(oldStatus, c.Status)
-	if err != nil {
-		panic(err)
-	}
-	if !validTrans {
-		err := fmt.Errorf("invalid claim item status transition from %s to %s",
-			oldStatus, c.Status)
+// Update changes the status if it is a valid transition.
+func (c *ClaimItem) Update(tx *pop.Connection, newStatus api.ClaimItemStatus, user User) error {
```
Baggerone, line 68: Did you need to change the oldStatus param to newStatus? That makes this function signature unique compared to  Item.Update() and Claim.Update().

I suppose you could change the others, if you prefer it this way.

application/models/claimitem.go
```diff
@@ -128,22 +133,22 @@ func claimItemStatusTransitions() map[api.ClaimItemStatus][]api.ClaimItemStatus
 	}
 }
 
-func isClaimItemTransitionValid(status1, status2 api.ClaimItemStatus) (bool, error) {
+func isClaimItemTransitionValid(status1, status2 api.ClaimItemStatus) bool {
```
Baggerone, line 136: It's nice getting rid of error return values.

Baggerone (approved):

application/models/claimitem.go
Schparky: I'm not too sure on that. It shouldn't be necessary since we have a hard restriction that this app will only work correctly if the OS is set to UTC. In fact, it might cause more harm than good to code it with `.UTC()`

application/models/claimitem.go
Schparky: Probably so. My brain was kind of fried by the end of the day yesterday.

application/models/claimitem.go
Schparky: Thanks

application/models/claimitem.go
Schparky: Yep. I just hope we don't find out the hard way that I was wrong about using `panic`.

application/api/claimitems.go
```diff
@@ -175,3 +175,16 @@ type ClaimFileAttachInput struct {
 	// swagger:strfmt uuid4
 	FileID uuid.UUID `json:"file_id"`
 }
+
+type ClaimItemUpdateInput struct {
+	// ClaimItem ID to update
+	//
+	// swagger:strfmt uuid4
+	ClaimItemID uuid.UUID `json:"claim_item_id"`
```
fillup, line 183: I don' think we need or want this right? The ID is in the URL already, so this would allow for the URL to specify one ID and then this model provide a different one.

Schparky: Actually, it's not. It's `PUT /claim-items/{id}`

Schparky: Ugh, you're right.

Schparky: I may have had that there when it was `PUT /claim/{id}/items` and forgot to remove it.

application/models/claimitem.go
fillup: I always add UTC() to be sure. I've had problems time and again when we don't use UTC on fixtures for access tokens (that is until updating them to use UTC). That might be more of a local dev issue if docker is getting time from the host and that's not in UTC, but I dunno. How about we just use UTC everywhere as a just-in-case-lets-make-sure-its-what-we-want kind of thing?

application/models/claimitem.go
fillup: Do you have to manually roll anything back yourself or just return an error and let the tx middleware handle it?

fillup (approved):
I left a couple small change comments, but otherwise looks good

application/models/claimitem.go
Schparky: OK. Since we always use UTC for production, that's fine.

application/models/claimitem.go
Schparky: Technically, no. I just hit test failures because the in-memory object was changed. Seems like it shouldn't change if something failed. Maybe I'm overthinking it.

Schparky: On further thought, that's correct. The return value should be ignored if `err != nil`.

application/actions/claimitems_test.go
Schparky: Right, the values are (or should be, at least) tested by the unit test(s) in the models package.

application/models/claimitem.go
Schparky: Right, there is an inconsistency here. I'll do another PR.

PR #108, opened 2021-09-03
A few claim emails

Baggerone:


Baggerone:
Will replace this with a bigger one that changes this one too.

PR #109, opened 2021-09-03
Claim emails

Baggerone:


application/listeners/claim.go
```diff
@@ -19,7 +39,24 @@ func claimSubmitted(e events.Event) {
 		return
 	}
 
-	// TODO Notify admin and do whatever else needs doing
+	if claim.Status != api.ClaimStatusReview1 {
+		domain.ErrLogger.Printf(wrongStatusMsg, "claimReview1", claim.Status)
```
Schparky, line 43: Should this return if the status isn't correct? Or, you could use `panic` since that will log the error message and end the listener goroutine.

application/listeners/claim.go
```diff
@@ -79,7 +167,23 @@ func claimReview2(e events.Event) {
 		return
 	}
 
-	// TODO Notify admin and do whatever else needs doing
+	if claim.Status != api.ClaimStatusReview2 {
+		domain.ErrLogger.Printf(wrongStatusMsg, "claimReview2", claim.Status)
```
Schparky, line 171: return or panic?

application/listeners/claim.go
```diff
@@ -94,7 +198,23 @@ func claimReview3(e events.Event) {
 		return
 	}
 
-	// TODO Notify admin boss and do whatever else needs doing
+	if claim.Status != api.ClaimStatusReview3 {
+		domain.ErrLogger.Printf(wrongStatusMsg, "claimReview3", claim.Status)
```
Schparky, line 202: return or panic?

application/listeners/claim_test.go
```diff
@@ -0,0 +1,373 @@
...
+
+	"github.com/gobuffalo/events"
+
+	"github.com/silinternational/cover-api/api"
+	"github.com/silinternational/cover-api/domain"
+	"github.com/silinternational/cover-api/models"
+	"github.com/silinternational/cover-api/notifications"
+)
+
+func (ts *TestSuite) Test_claimReview1() {
```
Schparky, line 14: It looks like it might be easy to combine all of these tests into one test with multiple subtests (or two tests, one for single-message and one multiple-message listener). The listener function could be a field of the test list struct.

application/listeners/item.go
```diff
@@ -14,7 +14,7 @@ import (
 const wrongStatusMsg = "error with %s listener. Item has wrong status: %s"
 
 func addMessageItemData(msg *notifications.Message, item models.Item) {
-	msg.Data["itemURL"] = fmt.Sprintf("%s/items/%s", domain.Env.UIURL, item.ID)
+	msg.Data["itemURL"] = fmt.Sprintf("%s/%s/%s", domain.Env.UIURL, domain.TypeItem, item.ID)
```
Schparky, line 17: I'm not sure this is a good fit, since the UI URL path and the API URL path don't necessarily have to be the same.

application/listeners/items_test.go
```diff
@@ -46,11 +46,8 @@ func (ts *TestSuite) Test_itemSubmitted() {
 		{
 			name: "just submitted, not approved",
 			event: events.Event{
-				Kind: domain.EventApiItemSubmitted,
-				Payload: events.Payload{
-					domain.EventPayloadID: submittedItem.ID,
-					EventPayloadNotifier:  &testEmailer,
-				},
+				Kind:    domain.EventApiItemSubmitted,
+				Payload: getTestPayload(submittedItem.ID, &testEmailer),
```
Schparky, line 50: This is a good extraction. I think the convention for a function that returns a new struct instance is `NewSomethingOrOther` so this could be `NewTestPayload`

Schparky (approved):

application/listeners/claim.go
Baggerone: :+1:

application/listeners/claim.go
Baggerone: :+1: panic

application/listeners/claim.go
Baggerone: :+1:

application/listeners/item.go
Baggerone: Oh, right.

jason-jackson (approved):

application/listeners/items_test.go
Baggerone: :+1:

application/listeners/claim_test.go
Baggerone: Well, some of the listeners will end up having other things to test besides the emails.
Also, I'm not sure about testing different functions in one set of tests. It sounds like that probably breaks some best practices.

I suppose maybe it could use a helper function or two. I'll look into that.

fillup (approved):

application/domain/domain.go
```diff
@@ -111,6 +111,13 @@ const (
 
 //  Notification templates
 const (
+	MessageTemplateClaimReview1Steward    = "claim_review1_steward"
```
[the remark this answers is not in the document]
Baggerone: Maybe to the new package in the next PR.

PR #110, opened 2021-09-03
start moving emails to messages

Baggerone:
This PR is now ready for a full review.

### DIFF
--- a/application/listeners/claim.go
+++ b/application/listeners/claim.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/silinternational/cover-api/api"
 	"github.com/silinternational/cover-api/domain"
+	"github.com/silinternational/cover-api/messages"
 	"github.com/silinternational/cover-api/models"
 	"github.com/silinternational/cover-api/notifications"
 )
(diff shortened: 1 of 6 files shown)


application/listeners/claim.go
```diff
@@ -43,20 +44,8 @@ func claimReview1(e events.Event) {
...
-	msg.Template = domain.MessageTemplateClaimReview1Steward
-	msg.Data["memberName"] = memberName
-	msg.Subject = "Action Required. " + memberName + " just (re)submitted a claim for approval"
-
 	notifiers := getNotifiersFromEventPayload(e.Payload)
-	if err := notifications.Send(msg, notifiers...); err != nil {
-		domain.ErrLogger.Printf("error sending claim review1 notification, %s", err)
-	}
-
+	messages.ClaimReview1(claim, notifiers)
```
jason-jackson, line 48: I think there should be an action. Either `messages.ClaimReview1(claim, notifiers).Send()` or `messages.SendClaimReview1(claim, notifiers)` or 
```
msg := messages.ClaimReview1(claim, notifiers)
notifications.SendMessage(msg)
```

application/messages/claim.go
```diff
@@ -0,0 +1,40 @@
...
+)
+
+func addMessageClaimData(msg *notifications.Message, claim models.Claim) {
+	msg.Data["claimURL"] = fmt.Sprintf("%s/%s/%s", domain.Env.UIURL, domain.TypeClaim, claim.ID)
+	msg.Data["claimRefNum"] = claim.ReferenceNumber
+	return
+}
+
+func newClaimMessageForMember(claim models.Claim, member models.User) notifications.Message {
+	msg := notifications.NewEmailMessage()
```
jason-jackson, line 18: Should this be here instead of in notifications?

Schparky (approved):

application/messages/claim.go
Baggerone: Yep. Good catch.

application/listeners/claim.go
Baggerone: Right. Steve had mentioned that too.

application/messages/claim.go
Baggerone: Actually. Probably not. 
`notifications` has its `Message` struct, which this is using.  So this function may as well stay with it, I think.

Schparky (approved):

PR #111, opened 2021-09-03
add a listener delay up front

Schparky:
This addresses a known issue that caused a panic in several listeners. Listeners were failing if a transaction that included an update did not close before the listener executed the database query.


Baggerone (approved):
Looks good.

fillup (approved):

PR #115, opened 2021-09-07
add accountable person and annual premium to items

Schparky:


application/models/item.go
```diff
@@ -403,3 +407,109 @@ func (i *Items) ConvertToAPI(tx *pop.Connection) api.Items {
...
+		if err := policyDependent.FindByID(tx, i.PolicyDependentID.UUID); err != nil {
+			panic("error finding policy dependent " + i.PolicyDependentID.UUID.String())
+		}
+		return policyDependent.Name
+	}
+	panic("no accountable person defined for item " + i.ID.String())
+}
+
+func (i *Item) GetAnnualPremium() int {
+	return (i.CoverageAmount + 25) / 50 // 2% rounded to the nearest cent
```
Baggerone, line 430: What is the `+ 25` for?

Baggerone (approved):

application/models/item.go
Schparky: To cause it to round up if the resulting fraction is greater than 0.5. Simply dividing by 50 will cause a truncation of the fractional part.

Baggerone: :+1:

application/api/users.go
```diff
@@ -30,6 +30,9 @@ type User struct {
 	// last name
 	LastName string `json:"last_name"`
 
+	// full name
+	Name string `json_"name"`
```
fillup, line 34:
```suggestion
	Name string `json:"name"`
```

PR #117, opened 2021-09-08
required fields for auto-approve

Baggerone:


application/models/item.go
```diff
@@ -265,11 +265,22 @@ func (i *Item) SubmitForApproval(tx *pop.Connection) error {
 	return nil
 }
 
+// Assumes the item already has its Category loaded
```
Schparky, line 268:
```suggestion
// isMissingRequiredFields checks for fields needed for auto-approval and returns `true` if any are not complete. Assumes the item already has its Category loaded
```

application/models/testutils.go
```diff
@@ -211,12 +211,23 @@ func createClaimFixture(tx *pop.Connection, policy Policy, config FixturesConfig
...
 // CreateCategoryFixtures generates any number of category records for testing
+//   even indexed categories are Stationary and odd indexed ones are Mobile
 func CreateCategoryFixtures(tx *pop.Connection, n int) Fixtures {
 	CreateRiskCategories(tx)
 
 	categories := make(ItemCategories, n)
+	even := true
 	for i := range categories {
-		categories[i].RiskCategoryID = RiskCategoryMobileID()
+		if even {
```
Schparky, line 221:
```suggestion
		if i % 2 == 0 {
```
I'm used to seeing this, but if you prefer the `even` variable, that's fine with me.

application/migrations/20210908222000_add_require_make_model_to_item_categories.up.fizz
```diff
@@ -0,0 +1,7 @@
+add_column("item_categories", "require_make_model", "bool", {"default": true})
+
+sql(`
+	UPDATE item_categories
+	SET require_make_model = false
+	WHERE risk_category_id = '7bed3c00-23cf-4282-b2b8-da89426cef2f';
```
Schparky, line 6: If `require_make_model` is determined by the risk category, is a separate field necessary?

Schparky (approved):

application/migrations/20210908222000_add_require_make_model_to_item_categories.up.fizz
Baggerone: On the spreadsheet, it looks like the `require_make_model` on the risk_category is only a default.

Schparky: So an `item_category` can be changed later to have a different value for `require_make_model`?

application/models/testutils.go
Baggerone: :+1:

Not that it would have mattered for the test fixture creation, but I was thinking the even-odd thing would be faster than `i%2==0`.
But, when I run a timed test for the two different methods (for a million iterations),  `i%2==0` is a little faster.

application/migrations/20210908222000_add_require_make_model_to_item_categories.up.fizz
Baggerone: @fillup What do you think about this?

fillup: I can't think of why that would have any connection to risk category. I wonder if that note is accurate in the spreadsheet. I'd think it would just be item category specific. I think we can probably drop the idea of the risk category having any awareness of make/model. If you want to wait a bit and check with Alex/Darryl to find out if there was a reason I can't imagine that's fine too.

PR #118, opened 2021-09-08
ensure accountable person ID is on the policy as a member or dependent

Schparky:


application/models/item_test.go
```diff
@@ -436,5 +436,66 @@ func (ms *ModelSuite) TestItem_LoadPolicyMembers() {
...
 
+	for _, tt := range tests {
+		ms.T().Run(tt.name, func(t *testing.T) {
+			err := tt.item.setAccountablePerson(ms.DB, tt.id)
+			if tt.appError != nil {
+				ms.Error(err, "test should have produced an error")
+				ms.EqualAppError(*tt.appError, err)
+				return
+			}
+			ms.NoError(err, "test produce an unexpected error")
```
Baggerone, line 490: *produced

Although, that makes it sound like you got a different error than the error you were expecting.
I'm thinking that no message is needed here, since the default ms.NoError message is basically the same (unless you want to give additional info).

Baggerone (approved):

application/models/item.go
```diff
@@ -491,25 +491,17 @@ func parseItemDates(input api.ItemInput, modelItem *Item) error {
 }
 
 func (i *Item) setAccountablePerson(tx *pop.Connection, id uuid.UUID) error {
-	var user User
-	err := user.FindByID(tx, id)
-	if err == nil {
+	i.LoadPolicy(tx, false)
+
+	if i.Policy.isMember(tx, id) {
 		i.PolicyUserID = nulls.NewUUID(id)
```
fillup, line 497: Would it make sense here, and below, to set the alternate field null to be sure only one or the other is set? I'm not sure how it'd be possible, but seems like it could be worth doing.

fillup (approved):

application/models/item.go
Schparky: Good idea

application/models/item_test.go
Schparky: Right

PR #124, opened 2021-09-09
Make sure the SameSite cookie warning isn't triggered

Baggerone:


Schparky (approved):
Good find

Baggerone:
Not working after all.

PR #125, opened 2021-09-09
create entity codes table

Schparky:


application/models/entitycode.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/gofrs/uuid"
+
+	"github.com/silinternational/cover-api/api"
+)
+
+type EntityCodes []EntityCode
+
+type EntityCode struct {
+	ID   uuid.UUID `db:"id"`
+	Code string    `db:"code"`
```
Baggerone, line 17: Do you want to include any validation?

application/models/entitycode.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	if perm == PermissionList {
+		return true
+	}
+
+	return false
+}
+
+func (ec *EntityCodes) ConvertToAPI(tx *pop.Connection) interface{} {
```
Baggerone, line 53: Does this need to be an interface return value?

Baggerone (approved):

fillup (approved):

application/models/entitycode.go
Schparky: Oops, I'll bet Goland did that when I used the "create method" helper.

application/models/entitycode.go
Schparky: Hmm, maybe not since it may never be a user-created data type.

PR #126, opened 2021-09-10
validate claimitem payout option vs claim event type

Baggerone:


application/api/claimitems.go
```diff
@@ -29,9 +29,10 @@ const (
 type PayoutOption string
 
 const (
-	PayoutOptionRepair      = PayoutOption("Repair")
-	PayoutOptionReplacement = PayoutOption("Replacement")
-	PayoutOptionFMV         = PayoutOption("FMV")
+	PayoutOptionRepair        = PayoutOption("Repair")
+	PayoutOptionReplacement   = PayoutOption("Replacement")
+	PayoutOptionFMV           = PayoutOption("FMV")
+	PayoutOptionFixedFraction = PayoutOption("FixedFraction")
```
Schparky, line 35: I realize this was settled this morning, but maybe "Partial" would convey the same meaning in fewer letters/words.

application/models/claimitem_test.go
```diff
@@ -82,11 +163,14 @@ func (ms *ModelSuite) TestClaimItem_Validate() {
 			if tt.wantErr {
 				if vErr.Count() == 0 {
 					t.Errorf("Expected an error, but did not get one")
+					return
```
Schparky, line 166: All of the if/else conditions are handled and there's nothing after the if, so I don't believe the returns are needed. Is it just for clarity?

Schparky (approved):

application/models/claimitem_test.go
Baggerone: Right. Sorry. It looked like it was not exiting the whole test on a failure, and that's what I was trying to accomplish.
I'll see if there is a good way to do that.

application/api/claimitems.go
Baggerone: For me "Partial" is a bit ambiguous.

PR #127, opened 2021-09-10
trying out alternate validation approach

fillup:


Schparky (approved):
Looks good, but there's one question I have. Since we have the `ClaimItem` field `IsRepairable` this may not work correctly. Not sure how that `IsRepairable` field is supposed to be used, though.

Baggerone (approved):
Looks good

PR #128, opened 2021-09-14
Getting started with queued notifications

Baggerone:
I see now I didn't need to send `notifiers...` through in my tests.
I'll back that out in a subsequent PR.

### DIFF
--- a/application/messages/item.go
+++ b/application/messages/item.go
@@ -2,6 +2,10 @@ package messages
 
 import (
 	"fmt"
+	"time"
+
+	"github.com/gobuffalo/nulls"
+	"github.com/gobuffalo/pop/v5"
 
 	"github.com/silinternational/cover-api/api"
 	"github.com/silinternational/cover-api/domain"
@@ -78,17 +82,43 @@ func ItemSubmittedSend(item models.Item, notifiers []interface{}) {
 	}
 }
 
-func ItemRevisionSend(item models.Item, notifiers []interface{}) {
+func ItemRevisionQueueMessage(tx *pop.Connection, item models.Item) {
 	item.LoadPolicyMembers(models.DB, false)
 
 	// TODO figure out how to specify required revisions
 
+	notn := models.Notification{
+		ItemID:  nulls.NewUUID(item.ID),
+		Subject: "changes have been requested on your new policy item",
+		// TODO make this more helpful
+		InappText: "changes have been requested on your new policy item",
+	}
+
+	data := newEmailMessageData()
+	data.addItemData(item)
+
 	for _, m := range item.Policy.Members {
-		msg := newItemMessageForMember(item, m)
-		msg.Template = MessageTemplateItemRevisionMember
-		msg.Subject = "changes have been requested on your new policy item"
-		if err := notifications.Send(msg, notifiers...); err != nil {
-			domain.ErrLogger.Printf("error sending item revision notification to member, %s", err)
+		data["memberName"] = m.Name()
+
+		notn2 := notn
+		notn2.Body = data.renderHTML(MessageTemplateItemRevisionMember)
+		// TODO Make these constants somewhere
+		notn2.Event = "Item Revision Required Notification"
+		notn2.EventCategory = "Item"
+
+		if err := notn2.Create(tx); err != nil {
+			panic("error creating new Notification: " + err.Error())
+		}
+
+		notnUser := models.NotificationUser{
+			NotificationID: notn2.ID,
+			UserID:         m.ID,
+			EmailAddress:   m.EmailOfChoice(),
+			SendAfterUTC:   time.Now().UTC(),
+		}
+
+		if err := notnUser.Create(tx); err != nil {
+			panic("error creating new NotificationUser: " + err.Error())
 		}
 	}
 }
--- a/application/messages/item.go
+++ b/application/messages/item.go
@@ -87,31 +87,27 @@ func ItemRevisionQueueMessage(tx *pop.Connection, item models.Item) {
 
 	// TODO figure out how to specify required revisions
 
-	notn := models.Notification{
-		ItemID:  nulls.NewUUID(item.ID),
-		Subject: "changes have been requested on your new policy item",
-		// TODO make this more helpful
-		InappText: "changes have been requested on your new policy item",
-	}
-
 	data := newEmailMessageData()
 	data.addItemData(item)
 
+	notn := models.Notification{
+		ItemID:  nulls.NewUUID(item.ID),
+		Body:    data.renderHTML(MessageTemplateItemRevisionMember),
+		Subject: "changes have been requested on your new policy item",
+		// TODO make this more helpful
+		InappText: "changes have been requested on your new policy item",
+
+		// TODO make these constants somewhere
+		Event:         "Item Revision Required Notification",
+		EventCategory: "Item",
+	}
+	if err := notn.Create(tx); err != nil {
+		panic("error creating new Notification: " + err.Error())
+	}
+
 	for _, m := range item.Policy.Members {
-		data["memberName"] = m.Name()
-
-		notn2 := notn
-		notn2.Body = data.renderHTML(MessageTemplateItemRevisionMember)
-		// TODO Make these constants somewhere
-		notn2.Event = "Item Revision Required Notification"
-		notn2.EventCategory = "Item"
-
-		if err := notn2.Create(tx); err != nil {
-			panic("error creating new Notification: " + err.Error())
-		}
-
 		notnUser := models.NotificationUser{
-			NotificationID: notn2.ID,
+			NotificationID: notn.ID,
 			UserID:         m.ID,
 			EmailAddress:   m.EmailOfChoice(),
 			SendAfterUTC:   time.Now().UTC(),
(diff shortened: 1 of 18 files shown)


application/messages/messages_test.go
```diff
@@ -53,3 +64,135 @@ func validateEmails(ts *TestSuite, td testData, testEmailer notifications.DummyE
...
+
+	type notnFixture struct {
+		name      string
+		sendAfter time.Time
+		sentAt    nulls.Time
+		notnUser  models.NotificationUser
+	}
+
+	// The database truncates the time in the fractions of seconds somehow
+	alreadySentTime := time.Now().UTC().Add(-domain.DurationWeek).Truncate(time.Second)
```
Schparky, line 108:
```suggestion
	alreadySentTime := time.Now().UTC().Add(-1 * domain.DurationWeek).Truncate(time.Second)
```
To make the negative more visible

application/messages/messages_test.go
```diff
@@ -53,3 +64,135 @@ func validateEmails(ts *TestSuite, td testData, testEmailer notifications.DummyE
...
+				"incorrect SentAtUTC value for AlreadySent")
+
+			ts.NoError(nuToSend.FindByID(db, nuToSend.ID),
+				"error fetching nuToSend from database")
+			ts.True(nuToSend.SentAtUTC.Valid, "missing SentAtUTC value")
+			ts.WithinDuration(time.Now().UTC(), nuToSend.SentAtUTC.Time, time.Minute,
+				"incorrect SentAtUTC value")
+
+			ts.NoError(nuNotReady.FindByID(db, nuNotReady.ID),
+				"error fetching nuToSend from database")
```
Schparky, line 185:
```suggestion
				"error fetching nuNotReady from database")
```

application/messages/messages_test.go
```diff
@@ -53,3 +64,135 @@ func validateEmails(ts *TestSuite, td testData, testEmailer notifications.DummyE
...
+			ts.False(nuNotReady.SentAtUTC.Valid, "SentAtUTC value should not be set")
+
+			ts.NoError(nuEmailError.FindByID(db, nuEmailError.ID),
+				"error fetching nuEmailError from database")
+			ts.False(nuEmailError.SentAtUTC.Valid, "SentAtUTC value should not be set")
+			ts.WithinDuration(time.Now().UTC().Add(time.Minute*10), nuEmailError.SendAfterUTC, time.Minute,
```
Schparky, line 191: Use a const or an Env for 10 minutes.

application/messages/messages.go
```diff
@@ -17,3 +31,62 @@ const (
...
+		msg := notifications.NewEmailMessage()
+		msg.ToName = n.User.Name()
+		msg.ToEmail = n.EmailAddress
+		msg.Body = n.Notification.Body
+		msg.Subject = n.Notification.Subject
+
+		if err := notifications.Send(msg); err != nil {
+			domain.ErrLogger.Printf("error sending queued notification email, %s", err)
+			// TODO figure out how to have a longer delay each time
+			n.SendAfterUTC = time.Now().UTC().Add(time.Minute * 10)
```
Schparky, line 82: Maybe add a const or even an Env for the 10 minute value.

application/models/notificationuser.go
```diff
@@ -0,0 +1,87 @@
...
+		}
+	}
+	if n.User.ID == uuid.Nil {
+		if err := tx.Load(n, "User"); err != nil {
+			panic("database error loading NotificationUser.User, " + err.Error())
+		}
+	}
+}
+
+func (n *NotificationUsers) GetQueuedEmails(tx *pop.Connection) error {
```
Schparky, line 73: Seems this should be named something like `GetEmailsToSend` since it doesn't return ALL queued emails.

application/models/validation.go
```diff
@@ -202,3 +202,24 @@ func policyStructLevelValidation(sl validator.StructLevel) {
...
+
+func notificationStructLevelValidation(sl validator.StructLevel) {
+	notn, ok := sl.Current().Interface().(Notification)
+	if !ok {
+		panic("notificationStructLevelValidation registered to a type other than Notification")
+	}
+
+	// Body and Subject are both required if InappText is blank or if either
+	// of them is present
+	if notn.InappText == "" || notn.Body != "" || notn.Subject != "" {
```
Schparky, line 214:
```suggestion
	if notn.InappText == "" {
```

application/models/validation.go
```diff
@@ -202,3 +202,24 @@ func policyStructLevelValidation(sl validator.StructLevel) {
...
+		if notn.Body == "" {
```
Schparky, line 215:
```suggestion
		if notn.Body == "" && notn.Subject != "" {
```

application/models/validation.go
```diff
@@ -202,3 +202,24 @@ func policyStructLevelValidation(sl validator.StructLevel) {
...
+			sl.ReportError(notn.Body, "body", "Body",
+				"body_required", "")
+
+		}
+		if notn.Subject == "" {
```
Schparky, line 220:
```suggestion
		if notn.Subject == "" && notn.Body != "" {
```

application/notifications/ses.go
```diff
@@ -28,14 +28,17 @@ type awsConfig struct {
...
 func (s *SES) Send(msg Message) error {
-	msg.Data["uiURL"] = domain.Env.UIURL
-	msg.Data["appName"] = domain.Env.AppName
-
-	bodyBuf := &bytes.Buffer{}
-	if err := eR.HTML(msg.Template).Render(bodyBuf, msg.Data); err != nil {
-		return errors.New("error rendering message body - " + err.Error())
+	body := msg.Body
+	if body == "" {
+		msg.Data["uiURL"] = domain.Env.UIURL
```
Schparky, line 33: Why is this is added here as well as in `newEmailMessageData`. Seems redundant.

Schparky (approved):

application/messages/messages_test.go
Baggerone: :+1:

application/messages/messages_test.go
Baggerone: :+1:

application/messages/messages_test.go
Baggerone: :+1:

application/messages/messages.go
Baggerone: :+1:

application/models/notificationuser.go
Baggerone: :+1:

application/models/validation.go
Baggerone: :+1:

application/notifications/ses.go
Baggerone: This is when the body still needs to be rendered.
newEmailMessageData is for situations when the body gets rendered before control passes to ses.

application/messages/item.go
```diff
@@ -78,17 +82,43 @@ func ItemSubmittedSend(item models.Item, notifiers []interface{}) {
...
+
 	for _, m := range item.Policy.Members {
-		msg := newItemMessageForMember(item, m)
-		msg.Template = MessageTemplateItemRevisionMember
-		msg.Subject = "changes have been requested on your new policy item"
-		if err := notifications.Send(msg, notifiers...); err != nil {
-			domain.ErrLogger.Printf("error sending item revision notification to member, %s", err)
+		data["memberName"] = m.Name()
+
+		notn2 := notn
```
fillup, line 103: I think we only need one notification record, but multiple notification user records, that's why it's two tables.

application/messages/item.go
```diff
@@ -78,17 +82,43 @@ func ItemSubmittedSend(item models.Item, notifiers []interface{}) {
...
+		notn2.Body = data.renderHTML(MessageTemplateItemRevisionMember)
+		// TODO Make these constants somewhere
+		notn2.Event = "Item Revision Required Notification"
+		notn2.EventCategory = "Item"
+
+		if err := notn2.Create(tx); err != nil {
+			panic("error creating new Notification: " + err.Error())
+		}
```
fillup, line 111: I think we only need one notification record, but multiple notification user records, that's why it's two tables. This can move outside the `for` loop then right?

application/messages/messages.go
fillup: Use an exponential backoff, something like:
```suggestion
			delay := n.SendAttemptCount^2 * time.Minute
			if n.SendAttemptCount > 10 {
				delay = 100 * time.Minute
			}
			n.SendAfterUTC = time.Now().UTC().Add(delay)
```

application/models/notificationuser.go
```diff
@@ -0,0 +1,87 @@
...
+	q := fmt.Sprintf(`SELECT notification_users.*
+  FROM notification_users LEFT JOIN notifications ON notification_users.notification_id = notifications.id
+  WHERE notifications.body <> '' AND
```
fillup, line 76: Do we need to check `notifications.body` here? Would we ever create a record with it being empty in the first place?

application/notifications/ses.go
fillup: Would that just be the case if the new notification model/process is not used?

fillup (approved):

application/messages/item.go
Baggerone: :+1:

PR #130, opened 2021-09-14
require a revision message on the ClaimsRequestRevision endpoint

Schparky:


application/models/claim.go
```diff
@@ -87,9 +87,10 @@ type Claim struct {
 	PaymentDate         nulls.Time            `db:"payment_date"`
 	TotalPayout         int                   `db:"total_payout"`
 	LegacyID            nulls.Int             `db:"legacy_id"`
+	RevisionMessage     string                `db:"revision_message" validate:"required_if=Status Revision"`
 	CreatedAt           time.Time             `db:"created_at"`
-	UpdatedAt           time.Time             `db:"updated_at"`
```
jason-jackson, line 91: Why was this moved?

jason-jackson (approved):

application/models/claim.go
Schparky: I trusted Goland too much. Used the "Create field RevisionMessage" and then the key for "Move Statement Up" to move it up a few lines. Didn't noticed that also moved the blank line.

application/models/claim.go
```diff
@@ -87,6 +87,7 @@ type Claim struct {
 	PaymentDate         nulls.Time            `db:"payment_date"`
 	TotalPayout         int                   `db:"total_payout"`
 	LegacyID            nulls.Int             `db:"legacy_id"`
+	RevisionMessage     string                `db:"revision_message" validate:"required_if=Status Revision"`
```
fillup, line 90: Hmm, do we want a single purpose field like this? I was thinking something like `StatusReason` so it could also apply to coverage `denied` too. It's okay to do it this way, but we may need to add something else to include a message in denials.

fillup (approved):

Baggerone (approved):

PR #132, opened 2021-09-16
a simple cache for the user object

Schparky:
In the work I'm doing on the policy history table, I found I wanted to re-load the current user when I only had the ID available. If this is too simplistic, I'll drop it. It does seem to work well, though.

### DIFF
--- a/application/models/user.go
+++ b/application/models/user.go
@@ -61,6 +61,8 @@ type User struct {
 	PhotoFile *File `belongs_to:"files"`
 }
 
+var userCache = map[uuid.UUID]User{}
+
 // Validate gets run every time you call a "pop.Validate*" (pop.ValidateAndSave, pop.ValidateAndCreate, pop.ValidateAndUpdate) method.
 //  It first adds a UUID to the user if its UUID is empty
 func (u *User) Validate(tx *pop.Connection) (*validate.Errors, error) {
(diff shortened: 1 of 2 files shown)


Baggerone (commented):
It seems like this is a bit risky. I don't know if it will benefit us enough to make it worth the risk.

I have a couple of concerns,
- there needs to be a way of updating the user object that's in the cache, when the user object in the database gets updated by any means (even through adminer).
- there needs to be a way to keep the cache from getting too big,
- I wonder if it will affect tests in unexpected ways.

Schparky:
> It seems like this is a bit risky. I don't know if it will benefit us enough to make it worth the risk.
> 
> I have a couple of concerns,
> 
> * there needs to be a way of updating the user object that's in the cache, when the user object in the database gets updated by any means (even through adminer).
> * there needs to be a way to keep the cache from getting too big,
> * I wonder if it will affect tests in unexpected ways.

I'm not going to dig my heels in, but my answers to your concerns are:

- This is pretty simple (except for the adminer part) if everything passes through `Update()`, which it should do anyway.
- I think it will never be too big since User has no child objects, but that's just an opinion
- Doubtful, unless we do queries and updates outside of the User methods

Baggerone:
What's the main benefit of using a cache?

fillup:
> It seems like this is a bit risky. I don't know if it will benefit us enough to make it worth the risk.
> 
> I have a couple of concerns,
> 
> * there needs to be a way of updating the user object that's in the cache, when the user object in the database gets updated by any means (even through adminer).
> * there needs to be a way to keep the cache from getting too big,
> * I wonder if it will affect tests in unexpected ways.

A couple thoughts:
1. Change the `userCache` struct to:
```go
type cachedUser struct{
  user User
  expires time.Time
}

type userCache map[uuid.UUID]cachedUser

func (c *userCache) Get(id uuid.UUID) *User {
  uc, ok := userCache[id]
  if !ok {
    return nil
  }
  if time.Now().UTC().After(uc.expires) {
    delete(userCache, id)
    return nil
  }
  
  return uc.User
}

func (c *userCache) Set(user User) {
  userCache[user.ID] = cachedUser{
    user: user,
    expires: time.Now().UTC().Add(60 * time.Minute)
  }
}
```

2. In the Update method that after a successful update call Set on the cache to create/update/overwrite the existing record? I wouldn't worry too much about the cache under the condition we updated something directly in db, adding this expiring capability will allow that to resolve itself. 

3. It'd be too much for this comment, but that cache thing could be extended to something like `recordCache` with property `record interface{}` instead of `user User`, and then we could have type specific getters, like `GetUser` that'd return the `record` cast to a `User`, but then we'd have the flexibility to cache other things over time if appropriate, like companies, categories, etc.

fillup:
> What's the main benefit of using a cache?

One benefit I see right away is not having to query db for every api call to find and authorize the user. of course if the user is load balanced across multiple servers we'd end up with difference caches on each of them, but we could also re-enable the sticky cookie to eliminate that.

Baggerone:
Is the query call slow enough to warrant avoiding it by complicating the code base?

Schparky:
> > What's the main benefit of using a cache?
> 
> One benefit I see right away is not having to query db for every api call to find and authorize the user. of course if the user is load balanced across multiple servers we'd end up with difference caches on each of them, but we could also re-enable the sticky cookie to eliminate that.

Which is probably a minute improvement given the database connection is not the largest contributor to latency. So it's somewhat of an academic exercise. Another benefit is it forces us to use User methods rather than going directly to the Pop library functions, bypassing any other logic we may have in place.

Schparky:
If there are no objections, I think I'll leave this PR open and revisit it after we're safely in sight of the 1.0 launch.

PR #133, opened 2021-09-16
policy history

Schparky:


application/models/policy.go
```diff
@@ -50,7 +52,21 @@ func (p *Policy) Create(tx *pop.Connection) error {
 }
 
 // Update writes the Policy data to an existing database record.
-func (p *Policy) Update(tx *pop.Connection) error {
+func (p *Policy) Update(ctx context.Context) error {
+	tx := Tx(ctx)
+	var oldPolicy Policy
+	if err := tx.Find(&oldPolicy, p.ID); err != nil {
+		return appErrorFromDB(err, api.ErrorUpdateFailure)
```
Baggerone, line 59: Should this be `api.ErrorQueryFailure` ?

application/models/policy.go
```diff
@@ -50,7 +52,21 @@ func (p *Policy) Create(tx *pop.Connection) error {
...
+	}
+
+	updates := p.Compare(oldPolicy)
+	for i := range updates {
+		history := p.NewHistory(ctx, api.HistoryActionUpdate, updates[i])
```
Baggerone, line 64: Maybe you have other ideas for using the Compare function, but if not, here's another idea. (Take it or leave it as you prefer)

You could create a function-as-a-variable inside the Compare function and then instead of appending to a slice and later looping through the slice, you could just call that function to create the history entries.

I suppose you would want to change the name of the function too.

application/models/policy_history.go
```diff
@@ -1,35 +1,64 @@
...
+	var user User
+	if err := user.FindByID(tx, p.UserID); err != nil {
+		panic("failed to find user by ID " + err.Error())
+	}
+
+	switch p.Action {
+	case api.HistoryActionCreate:
+		p.Description = fmt.Sprintf("record created by %s", user.Name())
+	case api.HistoryActionUpdate:
+		p.Description = fmt.Sprintf(`field %s changed from "%s" to "%s" by %s`,
```
Baggerone, line 58: Since `notes` could have a long value, maybe it would be better to have, `field %s changed by %s from "%s" to "%s"`

application/models/policy_test.go
```diff
@@ -171,3 +172,145 @@ func (ms *ModelSuite) TestPolicy_itemCoverageTotals() {
...
+				NewValue: newEntityCodeID,
+			},
+		},
+	}
+	for _, tt := range tests {
+		ms.T().Run(tt.name, func(t *testing.T) {
+			got := tt.policy.NewHistory(CreateTestContext(tt.user), api.HistoryActionUpdate, tt.update)
+			ms.False(tt.want.NewValue == tt.want.OldValue, "test isn't correctly checking a field update")
+			ms.Equal(tt.want.PolicyID, got.PolicyID, "PolicyID is not correct")
+			ms.Equal(tt.want.UserID, got.UserID, "FieldName is not correct")
```
Baggerone, line 308: `"UserID is not correct"`

Baggerone (approved):

application/models/policy_history.go
Schparky: Good point. And the column size for `description` is only 255. So I either need to increase it, or decide to not save changes to `notes`.

PR #134, opened 2021-09-16
return individual IDs (not the name of) the accountable person

Schparky:


application/api/items.go
```diff
@@ -90,8 +90,11 @@ type Item struct {
 	// annual premium (0.01 USD)
 	AnnualPremium int `json:"annual_premium"`
 
-	// Name of accountable person
-	AccountablePerson string `json:"accountable_person"`
+	// ID of a dependent designated as accountable person, must be null if policy_user_id is not null
+	PolicyDependentID nulls.UUID `json:"policy_dependent_id"`
+
+	// ID of a user designated as accountable person, must be null if policy_dependent_id is not null
```
Baggerone, line 96: Do these rules need to go in a validator?

Baggerone (approved):
Instead of `Policy` in their names, I wonder if it would be better to have `AccountableMemberID` and `AccountableDependentID`.

But, that's just an idea.  Feel free to leave it as you have it.

application/api/items.go
Schparky: Not necessary since the input struct only has one ID field, but I'll add it just for safety.

PR #135, opened 2021-09-16
Miscellaneous clean up

Baggerone:


application/actions/claimitems.go
```diff
@@ -8,7 +8,7 @@ import (
 	"github.com/silinternational/cover-api/models"
 )
 
-// swagger:operation PUT /claimitems/{id} ClaimItems ClaimItemsUpdate
+// swagger:operation PUT /claim-items/{id} ClaimItems ClaimItemsUpdate
```
Schparky, line 11: I'd have just committed this to develop so the UI could get it that much sooner

application/models/user.go
```diff
@@ -69,11 +69,17 @@ func (u *User) Validate(tx *pop.Connection) (*validate.Errors, error) {
 
 // Create stores the User data as a new record in the database.
 func (u *User) Create(tx *pop.Connection) error {
+	if u.AppRole == "" {
```
fillup, line 72: This is good, but can we also make sure when the call to create is happening during first login that the `AppRoleUser` role is explicitly set?

application/models/testutils.go
```diff
@@ -433,6 +433,10 @@ func DestroyAll() {
 	// delete all RiskCategories
 	var rCats RiskCategories
 	destroyTable(&rCats)
+
+	// delete all Notifications
+	var notfns Notifications
```
Schparky, line 438: You're pretty creative with your abbreviations. :)

Schparky (approved):
looks good to me, not overkill on the test cases

application/models/user.go
Baggerone: :+1:

application/models/testutils.go
Baggerone: I wasn't trying to be :-|   (I didn't want it to be `nots`.  I suppose I could have just used `ns`, since it is only used on the next line.)

PR #136, opened 2021-09-16
compact migration files

Schparky:
also removed the default risk category on items


application/migrations/20210802215427_create_policy_tables.up.fizz
```diff
@@ -1,14 +1,18 @@
...
 	t.Column("account", "string", {})
 	t.Column("entity_code", "string", {})
+	t.Column("legacy_id", "int", {"null": true})
+	t.Column("notes", "string", {"size": 2048})
+
 	t.Timestamps()
 
 	t.Index("type", {})
 	t.Index("household_id", {"unique": true})
+	t.Index("legacy_id", {"unique": true})
```
jason-jackson, line 15: Does this work? Won't it make it so you can't have more than one empty legacy_id?

application/migrations/20210803150753_create_item_categories.up.fizz
```diff
@@ -5,7 +5,11 @@ create_table("item_categories") {
 	t.Column("help_text", "string", {})
 	t.Column("status", "string", {})
 	t.Column("auto_approve_max", "integer", {})
+	t.Column("legacy_id", "int", {"null": true})
+
 	t.Timestamps()
 
+	t.Index("legacy_id", {"unique": true})
```
jason-jackson, line 12: Does this work? Won't it make it so you can't have more than one empty legacy_id?

For now it doesn't matter, but if we need new categories it may become an issue.

application/migrations/20210803202823_create_items.up.fizz
```diff
@@ -7,17 +7,25 @@ create_table("items") {
...
 	t.Column("purchase_date", "timestamp", {})
 	t.Column("coverage_status", "string", {})
 	t.Column("coverage_start_date", "timestamp", {})
+	t.Column("legacy_id", "int", {"null": true})
+	t.Column("risk_category_id", "uuid", {})
+
 	t.Timestamps()
 
+	t.Index(["name", "policy_id"], {"unique": true})
+	t.Index("legacy_id", {"unique": true})
```
jason-jackson, line 24: Does this work? Won't it make it so you can't have more than one empty legacy_id?

application/migrations/20210805195715_create_claims.up.fizz
```diff
@@ -9,9 +9,14 @@ create_table("claims") {
 	t.Column("reviewer_id", "uuid", {"null": true})
 	t.Column("payment_date", "timestamp", {"null": true})
 	t.Column("total_payout", "int", {})
+	t.Column("legacy_id", "int", {"null": true})
+	t.Column("reference_number", "string", {})
+
 	t.Timestamps()
 
 	t.Index("event_type", {})
+	t.Index("legacy_id", {"unique": true})
```
jason-jackson, line 18: Does this work? Won't it make it so you can't have more than one empty legacy_id?

jason-jackson (commented):
Most of my concern is with unique index for legacy_ids. I could be wrong as I'm more familiar with MySQL.

application/migrations/20210802215427_create_policy_tables.up.fizz
Schparky: No, because null is not considered in the unique index.

application/grifts/db.go
```diff
@@ -339,6 +339,7 @@ func createItemFixtures(tx *pop.Connection, fixPolicies []*models.Policy, fixICa
 			CoverageStatus:    api.ItemCoverageStatusApproved,
 			CoverageStartDate: time.Now().UTC().Add(time.Hour * time.Duration((i+1)*-40)),
 			PurchaseDate:      time.Now().UTC().Add(time.Hour * time.Duration((i+1)*-48)),
+			RiskCategoryID:    models.RiskCategoryStationaryID(),
```
Baggerone, line 342: Oops, I just added something like this in the PR I merged.

application/migrations/20210802215427_create_policy_tables.up.fizz
```diff
@@ -1,14 +1,18 @@
 create_table("policies") {
 	t.Column("id", "uuid", {primary: true})
 	t.Column("type", "string", {})
-	t.Column("household_id", "string", {})
+	t.Column("household_id", "string", {"null": true})
```
Baggerone, line 4: Is there a reason to make this different than cost_center and account (or to not make them {"null": true} as well)

Baggerone (approved):
Assuming you're planning to drop the tables in the staging database and start the migration from scratch.

(I didn't check the changes closely against the old migration files)

application/migrations/20210802215427_create_policy_tables.up.fizz
Schparky: It's only necessary here because of the unique key. I don't think the others can be unique.

application/grifts/db.go
Schparky: > Assuming you're planning to drop the tables in the staging database and start the migration from scratch.
> 
> (I didn't check the changes closely against the old migration files)

Wasn't planning to do that. I did a file compare on the schema.sql file before and after. The only differences are the column order and the removal of the risk_category default on the items table. Now that you mention it though, it's probably  a good idea to make sure there are absolutely no differences between development and staging.

application/migrations/20210803150753_create_item_categories.up.fizz
Schparky: No, because null is not considered in the unique index.

application/migrations/20210803202823_create_items.up.fizz
Schparky: No, because null is not considered in the unique index.

application/migrations/20210805195715_create_claims.up.fizz
Schparky: No, because null is not considered in the unique index.

PR #138, opened 2021-09-17
track claim history on update

Schparky:


fillup (approved):

application/api/api.go
```diff
@@ -148,3 +149,9 @@ func MergeExtras(extras []map[string]interface{}) map[string]interface{} {
 
 	return allExtras
 }
+
+type Currency int
+
+func (c Currency) String() string {
+	return fmt.Sprintf("%d.%d", c/domain.CurrencyFactor, c%domain.CurrencyFactor)
```
Baggerone, line 156: I think you need to figure out how to throw a `0` in there for the decimal if the decimal is below 10.

application/models/claim.go
```diff
@@ -112,14 +113,21 @@ func (c *Claim) Create(tx *pop.Connection) error {
...
-func (c *Claim) Update(tx *pop.Connection, oldStatus api.ClaimStatus) error {
-	validTrans, err := isClaimTransitionValid(oldStatus, c.Status)
+func (c *Claim) Update(ctx context.Context) error {
+	tx := Tx(ctx)
+
+	var oldClaim Claim
+	if err := tx.Find(&oldClaim, c.ID); err != nil {
+		return appErrorFromDB(err, api.ErrorQueryFailure)
+	}
+
```
Baggerone, line 123: Do we need to retain the oldStatus `StatusReason` for certain transitions?

Baggerone (approved):
Unless I'm mistaken, you definitely need to make one change before merging.

application/api/api.go
Schparky: Good catch. I'm only half awake today.

application/models/claim.go
Schparky: Not sure what you mean.

Baggerone: If the old Claim Status is Revision, we don't want the StatusReason to get changed do we?

Schparky: I don't think it will get changed. Only the input fields will change. It's not like `Item` where we use a covert-input-to-model function. Maybe I'm missing something.

Baggerone: OK :+1:

PR #139, opened 2021-09-17
more StatusReason work on Claims and other cleanup work

Baggerone:


application/messages/messages_test.go
```diff
@@ -58,14 +52,19 @@ func validateEmails(ts *TestSuite, td testData, testEmailer notifications.DummyE
 	ts.Len(msgs, wantCount, "incorrect message count")
 
 	gotTos := testEmailer.GetAllToAddresses()
-	ts.Equal(td.wantToEmails, gotTos)
+	ts.Equal(len(td.wantToEmails), len(gotTos), "incorrect count of ToEmail addresses")
```
Schparky, line 55: Not sure if there's a good reason to use it, but there's a `ts.Len` available.

Schparky (approved):
Nice cleanup of TODOs

PR #141, opened 2021-09-20
verify claim status before updating it on behalf of a user

Baggerone:
Checking if this is the right way to go before doing something similar on a ClaimItem update.

### DIFF
--- a/application/models/claim.go
+++ b/application/models/claim.go
@@ -153,6 +153,29 @@ func (c *Claim) Update(ctx context.Context) error {
 	return update(tx, c)
 }
 
+// UpdateByUser ensures the Claim has an appropriate status for being modified by the user
+//  and then writes the Claim data to an existing database record.
+func (c *Claim) UpdateByUser(ctx context.Context) error {
+	user := CurrentUser(ctx)
+	if user.IsAdmin() {
+		return c.Update(ctx)
+	}
+
+	switch c.Status {
+	case api.ClaimStatusDraft, api.ClaimStatusRevision, api.ClaimStatusReview1:
+	default:
+		err := errors.New("user may not edit a claim that is too far along in the review process.")
+		appErr := api.NewAppError(err, api.ErrorClaimStatus, api.CategoryUser)
+		return appErr
+	}
+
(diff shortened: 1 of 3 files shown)


application/models/claim.go
```diff
@@ -153,6 +153,29 @@ func (c *Claim) Update(ctx context.Context) error {
...
+// UpdateByUser ensures the Claim has an appropriate status for being modified by the user
+//  and then writes the Claim data to an existing database record.
+func (c *Claim) UpdateByUser(ctx context.Context) error {
+	user := CurrentUser(ctx)
+	if user.IsAdmin() {
+		return c.Update(ctx)
+	}
+
+	switch c.Status {
+	case api.ClaimStatusDraft, api.ClaimStatusRevision, api.ClaimStatusReview1:
```
Schparky, line 165: Nice, I like this better than an `if`, even though there is only one `case`.

application/models/claim.go
```diff
@@ -153,6 +153,29 @@ func (c *Claim) Update(ctx context.Context) error {
...
+	default:
+		err := errors.New("user may not edit a claim that is too far along in the review process.")
+		appErr := api.NewAppError(err, api.ErrorClaimStatus, api.CategoryUser)
+		return appErr
+	}
+
+	if c.Status == api.ClaimStatusReview1 {
```
Schparky, line 172: I feel like this deserves a comment. Why is it changing the status? I know the answer because it was recently discussed, but that will probably fade.

Schparky (approved):

application/models/claim.go
Baggerone: Yeah, it saves repeating the first part (`c.Status == ...`)

fillup: maybe add a comment in this case that you're leaving it blank to explicitly do nothing when the user is allowed to make the change.

fillup (approved):

application/models/claim.go
Baggerone: :+1:

application/models/claim.go
Baggerone: :+1:

PR #145, opened 2021-09-20
ClaimItem update

Baggerone:
I simplified the logic for setting the Reviewer fields on the ClaimItem.  It didn't seem like precision was too important there but let me know if you disagree.

### DIFF
--- a/application/models/claimitem.go
+++ b/application/models/claimitem.go
@@ -1,6 +1,8 @@
 package models
 
 import (
+	"context"
+	"errors"
 	"fmt"
 	"net/http"
 	"time"
@@ -16,8 +18,11 @@ import (
 
 var ValidClaimItemStatus = map[api.ClaimItemStatus]struct{}{
 	api.ClaimItemStatusDraft:    {},
-	api.ClaimItemStatusPending:  {},
+	api.ClaimItemStatusReview1:  {},
+	api.ClaimItemStatusReceipt:  {},
 	api.ClaimItemStatusRevision: {},
+	api.ClaimItemStatusReview2:  {},
+	api.ClaimItemStatusReview3:  {},
 	api.ClaimItemStatusApproved: {},
 	api.ClaimItemStatusDenied:   {},
 }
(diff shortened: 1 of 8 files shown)


application/models/claimitem.go
```diff
@@ -64,24 +68,68 @@ func (c *ClaimItem) Validate(tx *pop.Connection) (*validate.Errors, error) {
...
+
+	// Maybe One day we will want to worry about the status on the ClaimItem itself
+	//if !isClaimItemTransitionValid(oldStatus, c.Status) {
+	//	err := fmt.Errorf("invalid claim item status transition from %s to %s", oldStatus, c.Status)
+	//	appErr := api.NewAppError(err, api.ErrorValidation, api.CategoryUser)
+	//	return appErr
+	//}
+
+	// Set the Reviewer fields when needed.
+	if user.IsAdmin() {
```
Schparky, line 93: Originally, the reviewer on the Claim Item was going to be the Steward and the reviewer on the Claim was to be the Signator. But I think things have changed. We may want to add a second reviewer field on the Claim and get rid of this one.

application/models/claimitem.go
```diff
@@ -115,37 +163,14 @@ func (c *ClaimItem) IsActorAllowedTo(tx *pop.Connection, actor User, perm Permis
...
-		},
-		api.ClaimItemStatusRevision: {
-			api.ClaimItemStatusPending,
-		},
-		api.ClaimItemStatusApproved: {},
-		api.ClaimItemStatusDenied:   {},
-	}
-}
-
 func isClaimItemTransitionValid(status1, status2 api.ClaimItemStatus) bool {
```
Schparky, line 166: This function isn't being called, but it is updated. Did you intend to update the calling code (instead of commenting it)?

Schparky (approved):

application/models/claimitem.go
Baggerone: If you don't mind, I think I'll leave this here for this PR and maybe some day we can decide exactly how we want to deal with it.

PR #146, opened 2021-09-21
Policy invites

Baggerone:
There's more to do about accepting the invite, but I'll leave that for another PR.

Also, since were using queued notifications, I didn't do anything with the two fields `EmailSentAt` and `EmailSendCount` on the invites.
Should I do something with them?  Now or later?


application/models/notificationuser.go
```diff
@@ -18,7 +18,8 @@ type NotificationUsers []NotificationUser
 type NotificationUser struct {
 	ID               uuid.UUID  `db:"id"`
 	NotificationID   uuid.UUID  `db:"notification_id"`
-	UserID           uuid.UUID  `db:"user_id"`
+	UserID           nulls.UUID `db:"user_id"`
```
Schparky, line 21: I searched for usage of `UserID` and `User`. Seems they aren't needed. Is that true?

Schparky (approved):

application/models/notificationuser.go
Baggerone: UserID is used in messages.SendQueuedNotifications()

application/actions/policies.go
```diff
@@ -139,6 +142,75 @@ func policiesListMembers(c buffalo.Context) error {
...
+
+	policy.LoadMembers(tx, false)
+
+	var invite api.PolicyUserInviteCreate
+	if err := StrictBind(c, &invite); err != nil {
+		return reportError(c, err)
+	}
+
+	// make sure user is not already a member of this policy
+	for _, m := range policy.Members {
```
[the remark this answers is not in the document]
Baggerone: Good catch `func (p *Policy) isMember`

application/models/notificationuser.go
Schparky: True, but only to get the name, which is redundant. ... unless I'm still missing it

Schparky: Maybe the purpose is for future use...to track which user was invited when and by whom

Baggerone: Well, notificationUsers were originally for user objects from the database, but the policy invites are for future users (not in the db). So, I was trying to cover for both cases.

Baggerone: As you say, I think it will probably be helpful in the future to have the link to the db User for other purposes.

fillup: In the future when we have in app notifications visible we can use the user_id to get the current user's list of notifications.

PR #150, opened 2021-09-22
Accept invite

Baggerone:
This still needs some manual testing once the UI is hooked up to use it.


jason-jackson:
I'm curious if it's possible to check email as well.

application/models/policyuserinvite.go
```diff
@@ -105,3 +118,58 @@ func (i *PolicyUserInvite) LoadPolicy(tx *pop.Connection, reload bool) {
...
+func (i *PolicyUserInvite) DestroyIfExpired(tx *pop.Connection) error {
+	now := time.Now().UTC()
+	cutoff := now.Add(time.Duration(-domain.Env.InviteLifetimeDays) * domain.DurationDay)
+	if i.CreatedAt.After(cutoff) {
+		return nil
+	}
+
+	if err := tx.Destroy(i); err != nil {
+		return api.NewAppError(
+			errors.New("error destroying expired invite: "+err.Error()),
```
Schparky, line 164: `appErrorFromDB` would work, but you'd lose the "expired" word, so it's a tradeoff. The advantage to `appErrorFromDB` is that it gets the detailed error from the database driver that is helpful for troubleshooting.

Schparky (approved):

application/models/policyuserinvite.go
Schparky: To get the best of both, you could assign the result of `appErrorFromDB` to a variable and then add to the message or use `fmt.Errorf("error destroying expired invite %w", err)` before returning it.

Baggerone:
I think we could verify the email address against the one on the invite.  I'm not sure if we want to be that strict, since outsiders won't be able to log in to the IDP anyway.

If we were to require a match on the email address, maybe we would need to validate the email before an invite is created to ensure it matches our domains.

application/models/policyuserinvite.go
```diff
@@ -85,6 +87,17 @@ func (i *PolicyUserInvite) Save(tx *pop.Connection) error {
 	return save(tx, i)
 }
 
+func (i *PolicyUserInvite) Destroy(tx *pop.Connection) error {
+	if err := tx.Destroy(i); err != nil {
+		return api.NewAppError(
```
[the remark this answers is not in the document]
Baggerone: :+1:

jason-jackson:
> I think we could verify the email address against the one on the invite. I'm not sure if we want to be that strict, since outsiders won't be able to log in to the IDP anyway.
> 
> If we were to require a match on the email address, maybe we would need to validate the email before an invite is created to ensure it matches our domains.

I think if we decide to validate email, we could probably delay it until after 1.0.

jason-jackson (approved):

PR #155, opened 2021-09-29
mostly just some import stuff

Schparky:


application/grifts/import.go
```diff
@@ -450,18 +452,27 @@ func normalizePolicy(p *LegacyPolicy) error {
 		p.EntityCode = nulls.String{}
 
 		if p.HouseholdId == "" {
-			return fmt.Errorf("Policy[%s] HouseholdId is empty, dropping policy", p.Id)
+			randomNumber := domain.RandomString(20, "0123456789")
```
jason-jackson, line 455: Should anything be done to ensure random number doesn't already exist?

application/grifts/import.go
```diff
@@ -450,18 +452,27 @@ func normalizePolicy(p *LegacyPolicy) error {
...
+			log.Printf("Policy[%s] HouseholdId is empty, using random number: %s", p.Id, randomNumber)
 		}
 	}
 
 	if p.Type == "ou" || p.Type == "corporate" {
 		p.HouseholdId = ""
 
 		if !p.EntityCode.Valid || p.EntityCode.String == "" {
-			return fmt.Errorf("Policy[%s] EntityCode is empty, dropping policy", p.Id)
+			randomNumber := domain.RandomString(20, "0123456789")
```
jason-jackson, line 466: Should anything be done to ensure random number doesn't already exist?

application/grifts/import.go
```diff
@@ -450,18 +452,27 @@ func normalizePolicy(p *LegacyPolicy) error {
...
+			p.EntityCode = nulls.NewString(randomNumber)
+			p.Notes += fmt.Sprintf(" [autogenerated EntityCode %s during migration]", randomNumber)
+			log.Printf("Policy[%s] EntityCode is empty, using random number: %s", p.Id, randomNumber)
 		}
 		if p.CostCenter == "" {
-			return fmt.Errorf("Policy[%s] CostCenter is empty, dropping policy", p.Id)
+			randomNumber := domain.RandomString(20, "0123456789")
```
jason-jackson, line 472: Should anything be done to ensure random number doesn't already exist?

Baggerone, line 472: I wonder if it would better to use a constant here that is easy to identify (e.g. all 9's).

application/grifts/import.go
```diff
@@ -450,18 +452,27 @@ func normalizePolicy(p *LegacyPolicy) error {
...
 		}
 	}
 
 	if p.Type == "ou" || p.Type == "corporate" {
 		p.HouseholdId = ""
 
 		if !p.EntityCode.Valid || p.EntityCode.String == "" {
-			return fmt.Errorf("Policy[%s] EntityCode is empty, dropping policy", p.Id)
+			randomNumber := domain.RandomString(20, "0123456789")
+			p.EntityCode = nulls.NewString(randomNumber)
```
Baggerone, line 467: I wonder if it would better to use a constant here that is easy to identify (e.g. all 9's).

application/grifts/import.go
```diff
@@ -859,7 +876,7 @@ func formatDate(d string) string {
 }
 
 func parseStringTime(t, desc string) time.Time {
-	if t == "" {
+	if t == "" || strings.Contains(t, "1970-01-01") {
```
Baggerone, line 879: The date string could be a const.

application/grifts/import.go
```diff
@@ -892,7 +909,7 @@ func parseNullStringTimeToTime(t nulls.String, desc string) time.Time {
 }
 
 func parseStringTimeToNullTime(t, desc string) nulls.Time {
-	if t == "" {
+	if t == "" || strings.Contains(t, "1970-01-01") {
```
Baggerone, line 912: The date string could be a const

application/models/claim.go
```diff
@@ -158,6 +158,9 @@ func (c *Claim) Update(ctx context.Context) error {
 func (c *Claim) UpdateByUser(ctx context.Context) error {
 	user := CurrentUser(ctx)
 	if user.IsAdmin() {
+		if c.Status.IsReview() {
```
Baggerone, line 161: I'm thinking that this doesn't belong here. At times, the steward may do an action on behalf of the customer when it's not a review type of step.

application/models/claim.go
```diff
@@ -337,6 +340,7 @@ func (c *Claim) SubmitForApproval(ctx context.Context) error {
 		err := fmt.Errorf("invalid claim status for submit: %s", oldStatus)
 		return api.NewAppError(err, api.ErrorClaimStatus, api.CategoryUser)
 	}
+	c.setReviewer(ctx)
```
Baggerone, line 343: Same comment about this not being needed here.

application/grifts/import.go
Schparky: In "normal" or production code, I'd say so. But this can easily be re-executed in the rare case where a duplicate is generated.

application/grifts/import.go
Schparky: Normally I like to use constants, but this string is very self-documenting in my opinion. What would you suggest as a const name for this? `allNumericDigits`, `allNumbers`?

application/grifts/import.go
Schparky: OK

application/models/claim.go
Schparky: So we should remove the validation that requires the ReviewerID and ReviewDate on the applicable statuses?

application/grifts/import.go
Baggerone: I was referring more to choosing one particular value for all the blank cost centers rather than using a different random number for each.

application/models/claim.go
jason-jackson: I thought we wanted to stay away from allowing steward to perform actions on behalf of a customer?

Schparky: OK, I'll remove the validation. @jason-jackson I think that in the short term, at least, the steward will need to continue to "manage" policies until users take control for themselves.

application/grifts/import.go
Schparky: In that case, I think all three can be a placeholder value. There's only one remaining policy with missing household ID so the unique key isn't a problem.

Baggerone (approved):

PR #156, opened 2021-09-29
Avoid error when removing an item associated with a LedgerEntry

Baggerone:


application/models/item.go
```diff
@@ -95,25 +95,26 @@ func (i *Item) FindByID(tx *pop.Connection, id uuid.UUID) error {
...
-
-	if clICount > 0 {
-		return i.Inactivate(tx)
+	// Just inactivate (don't delete) if there are associated claim items or
+	//  if there are associated ledger entries
+	relations := []interface{}{&ClaimItems{}, &LedgerEntries{}}
+	for j, _ := range relations {
+		mCount, err := tx.Where("item_id = ?", i.ID).Count(relations[j])
+		if err != nil {
+			return api.NewAppError(
```
Schparky, line 111: Would `AppErrorFromDB` work here?

Schparky (approved):
This looks good. I wonder, though, would it be good enough to just base it on the status? There should be no ledger entries for items that are not yet approved.

Schparky (approved):

PR #159, opened 2021-09-30
initial work on getting only the active items for a policy

Baggerone:
Not intended for merging


Baggerone:
Not needed for now at least.

PR #160, opened 2021-09-30
export transaction batch to CSV for import into Sage

Schparky:


application/models/item.go
```diff
@@ -530,15 +540,65 @@ func (i *Item) setAccountablePerson(tx *pop.Connection, id uuid.UUID) error {
...
 	return le.Create(tx)
 }
+
+// getIncomeAccount maps the item data to the income account for billing
+//
+// WARNING: requires Policy and Policy.EntityCode to be pre-loaded
+func (i *Item) getIncomeAccount(tx *pop.Connection) string {
+	// TODO: move hard-coded account numbers to the database or to environment variables
+	accountMap := map[string]string{
+		"MMB/STM": "40200",
```
Baggerone, line 566: Same point about potentially sensitive info.

application/models/item.go
```diff
@@ -530,15 +540,65 @@ func (i *Item) setAccountablePerson(tx *pop.Connection, id uuid.UUID) error {
...
+func (i *Item) GetAccountablePersonName(tx *pop.Connection) (firstName, lastName string) {
+	if i.PolicyUserID.Valid {
+		var user User
+		_ = user.FindByID(tx, i.PolicyUserID.UUID)
+		return user.FirstName, user.LastName
+	}
+	if i.PolicyDependentID.Valid {
+		var dep PolicyDependent
+		_ = dep.FindByID(tx, i.PolicyDependentID.UUID)
+		names := strings.SplitN(dep.Name, " ", 2)
```
Baggerone, line 597: Does it matter if the split is wrong, e.g. with "Mary Ann Nelson", it would make "Ann Nelson" the last name, right?
I guess if you try to do it the other way, it would make certain other names wrong.

application/models/item.go
```diff
@@ -530,15 +540,65 @@ func (i *Item) setAccountablePerson(tx *pop.Connection, id uuid.UUID) error {
...
+	}
+	i.LoadRiskCategory(tx, false)
+	incomeAccount := accountMap[billingEntity] + i.RiskCategory.CostCenter
+
+	return incomeAccount
+}
+
+// GetAccountablePersonName gets the name of the accountable person. In case of error, empty strings
+// are returned.
+func (i *Item) GetAccountablePersonName(tx *pop.Connection) (firstName, lastName string) {
```
Baggerone, line 588: I'm not so sure about returning a mix of named params and user attributes. It seems like just having explicit return statements is better.

I guess if you really want them named for documentation purposes, you could still include them in the return statement at the end.

application/models/item_test.go
```diff
@@ -718,17 +718,35 @@ func (ms *ModelSuite) TestItem_GetProratedPremium() {
...
+	item0 := f.Items[0]
+	ms.NoError(item0.setAccountablePerson(ms.DB, f.Users[0].ID))
+	first, last := item0.GetAccountablePersonName(ms.DB)
+	ms.Equal(f.Users[0].FirstName, first, "first name is not correct")
+	ms.Equal(f.Users[0].LastName, last, "last name is not correct")
+
+	item1 := f.Items[1]
+	ms.NoError(item1.setAccountablePerson(ms.DB, f.PolicyDependents[0].ID))
+	first, last = item1.GetAccountablePersonName(ms.DB)
+	ms.Contains(f.PolicyDependents[0].Name, first, "first name is not correct")
```
Baggerone, line 750: I'm thinking that the way the name is created in the fixtures, these two lines aren't so helpful.

application/models/ledgerentry.go
```diff
@@ -1,41 +1,154 @@
...
+
+	"github.com/silinternational/cover-api/api"
+	"github.com/silinternational/cover-api/domain"
+	"github.com/silinternational/cover-api/fin"
+)
+
+type LedgerEntryRecordType string
+
+const (
+	LedgerEntryRecordTypeNewCoverage      = LedgerEntryRecordType("NewCoverage")
```
Baggerone, line 19: It would be nice if we didn't need `Record` in these names. Is it really needed?

application/models/ledgerentry.go
```diff
@@ -1,41 +1,154 @@
...
-func (ec *LedgerEntry) Create(tx *pop.Connection) error {
-	return create(tx, ec)
+func (le *LedgerEntry) Create(tx *pop.Connection) error {
+	return create(tx, le)
+}
+
+func (le *LedgerEntries) AllForMonth(tx *pop.Connection, firstDay time.Time) error {
+	lastDay := domain.EndOfMonth(firstDay)
+
+	err := tx.Where("date_submitted BETWEEN ? and ?", firstDay, lastDay).
```
Baggerone, line 73: I don't know the answer ... will this include the ledger entries from the last day of the month?

application/models/ledgerentry.go
```diff
@@ -1,41 +1,154 @@
...
+		Where("date_entered IS NULL").All(le)
+
+	return appErrorFromDB(err, api.ErrorQueryFailure)
+}
+
+type TransactionBlocks map[string]LedgerEntries // keyed by account
+
+func (le *LedgerEntries) ToCsv(batchDate time.Time) []byte {
+	sage := fin.NewBatch("sage", batchDate)
```
Baggerone, line 82: Can "sage" be a constant somewhere?

application/fin/fin.go
```diff
@@ -0,0 +1,40 @@
...
+type Provider interface {
+	AppendToBatch(Transaction)
+	BatchToCSV() []byte
+}
+
+func NewBatch(providerType string, date time.Time) Provider {
+	batchDesc := fmt.Sprintf("%s %s JE", date.Format("January 2006"), domain.Env.AppName)
+
+	switch providerType {
+	case "sage":
```
Baggerone, line 27: Can "sage" be a constant somewhere?

application/actions/app.go
```diff
@@ -139,6 +140,11 @@ func App() *buffalo.App {
 		auth.POST("/callback", authCallback)
 		auth.GET("/logout", authDestroy)
 
+		// accounting batches
+		batchesGroup := app.Group(batchesPath)
+		batchesGroup.Middleware.Skip(AuthZ, batchesGetLatest)
```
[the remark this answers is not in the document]
Schparky: Yes, I should add a TODO here.

application/actions/batches.go
```diff
@@ -0,0 +1,50 @@
...
+
+	if len(le) == 0 {
+		return c.Render(http.StatusNoContent, nil)
+	}
+
+	csvData := le.ToCsv(firstDay)
+
+	response := c.Response()
+	response.Header().Set("Content-Type", "text/csv")
+	response.Header().Set("Content-Disposition", `attachment; filename="batch.csv"`)
```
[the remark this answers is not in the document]
Schparky: Good catch

application/domain/domain.go
```diff
@@ -172,10 +172,13 @@ var Env struct {
 	// The following will be multiplied by CurrencyFactor in readEnv()
 	PolicyMaxCoverage       int `default:"50000" split_words:"true"`
 	DependentAutoApproveMax int `default:"4000" split_words:"true"`
-	PremiumMinimum          int `default:"25"`
+	PremiumMinimum          int `default:"25" split_words:"true"`
 
 	// PremiumFactor is multiplied by CoverageAmount to calculate the annual premium of an item
 	PremiumFactor float64 `default:"0.02"`
```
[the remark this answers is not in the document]
Schparky: Yep. Don't know how I missed that one

application/fin/sage.go
```diff
@@ -0,0 +1,60 @@
...
+
+import (
+	"bytes"
+	"fmt"
+
+	"github.com/silinternational/cover-api/api"
+)
+
+const (
+	header1 = `"RECTYPE","BATCHID","BTCHENTRY","ORIGCOMP","SRCELEDGER","SRCETYPE","FSCSYR","FSCSPERD","SWEDIT","JRNLDESC","REVPERD","ERRBATCH","ERRENTRY","DETAILCNT","PROCESSCMD"` + "\n"
```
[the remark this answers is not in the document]
Schparky: Sure. Didn't think anyone would care to read that gibberish anyway. :-)

application/domain/domain.go
```diff
@@ -399,3 +402,11 @@ func CalculatePartialYearValue(value int, startDate time.Time) int {
 
 	return value * daysSince / 365
 }
+
+func BeginningOfLastMonth(date time.Time) time.Time {
```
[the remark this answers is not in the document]
Schparky: OK. I'm planning to work on tests today. Wanted to get a PR out before it grew even bigger. On that subject, do you find that having tests makes the rest of the code easier to read?

application/grifts/import.go
```diff
@@ -857,6 +854,48 @@ func importJournalEntries(tx *pop.Connection, entries []JournalEntry) {
 	fmt.Printf("  LedgerEntries: %d (policy IDs not found: %s)\n", nImported, strings.Join(s, ","))
 }
 
+func getIncomeAccount(e JournalEntry) string {
+	accountMap := map[string]string{
+		"MMB/STM": "40200",
```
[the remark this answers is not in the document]
Schparky: If they're sensitive, it's too late now. Oops. Don't think so, though. For import code, I expect to remove it once we do the migration. For app code, I plan to move it all into environment variables or database records.

application/migrations/20210928212512_ledger_changes_for_financial_reporting.down.fizz
```diff
@@ -0,0 +1,10 @@
+drop_column("risk_categories")
```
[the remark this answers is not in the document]
Schparky: Yep, thanks. I usually test the down migration but must have forgotten to do that this time.

application/models/item.go
Schparky: Same answer. ;)

application/models/item.go
Schparky: Honestly, I was puzzled to see that we use split names for users but combined for dependents. I think that should be changed.

application/models/item.go
Schparky: I think it's in line with [Effective Go](https://golang.org/doc/effective_go#named-results) and [Code Review Comments](https://golang.org/doc/effective_go#named-results). Since they are both the same type, the named returns are important. Since "a handful of lines" is pretty subjective, I'll go ahead and add explicit names on the return.

application/models/item_test.go
Schparky: Do you mean that since the dependent name has no space? I'll add a space in there, or as I mentioned I think it ought to be consistent with the User model and have two separate columns/fields. Maybe I'll just do that.

application/models/ledgerentry.go
Schparky: I just about dropped it. I guess because the field name is `RecordType` I thought it was necessary. Should the field name be changed to just `Type`?

application/models/ledgerentry.go
Schparky: Yep.

application/models/ledgerentry.go
Schparky: Yep

application/fin/fin.go
Schparky: Yep

application/models/ledgerentry.go
Baggerone: I suppose the field could just be `Type` if you think that doesn't cause a problem somewhere.

application/domain/domain.go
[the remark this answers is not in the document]
Baggerone: In some cases, having a test helps you see what the code accomplishes. Also, sometimes, if you're not sure the code will do what it was intended to do, having a test there might put your mind at ease.

Like with this date arithmetic, the Go documentation says that AddDate does funny things sometimes, e.g. adding a month to Oct. 31 gives you Dec. 1 (if I understood that correctly).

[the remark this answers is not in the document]
Schparky: I totally agree. I was asking in the context of PR review. Does not having a test make the code harder to read during a PR review? I know some people say that tests are a form of documentation. Not sure I agree, which is why I asked.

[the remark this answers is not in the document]
Baggerone: I suppose it depends on the situation.  If a function is fairly clear, then having a test may not help the review. If it's mysterious somehow, then a test would keep me from feeling like I need to use the Go playground to double check that it's doing the right thing.

application/models/item_test.go
Baggerone: :+1:

application/models/ledgerentry.go
```diff
@@ -1,41 +1,154 @@
...
+	return create(tx, le)
+}
+
+func (le *LedgerEntries) AllForMonth(tx *pop.Connection, firstDay time.Time) error {
+	lastDay := domain.EndOfMonth(firstDay)
+
+	err := tx.Where("date_submitted BETWEEN ? and ?", firstDay, lastDay).
+		Where("date_entered IS NULL").All(le)
+
+	return appErrorFromDB(err, api.ErrorQueryFailure)
```
Baggerone, line 76: I just noticed this. (I was stealing your query idea.)
It looks like it's assuming there will definitely be an error.

Schparky: Just making sure I understand your comment. Are you saying the code is hard to read, or you're concerned there's a problem?

Baggerone: Shouldn't there be an `if` in there to check that there really is an error?

Schparky: There is one inside `appErrorFromDB`.

application/grifts/import.go
```diff
@@ -857,6 +854,48 @@ func importJournalEntries(tx *pop.Connection, entries []JournalEntry) {
...
+
+	if e.PolicyType == 2 {
+		incomeAccount += "MPRO12"
+	} else {
+		incomeAccount += "MCMC12"
+	}
+	return incomeAccount
+}
+
+func getLedgerRecordType(i int) models.LedgerEntryType {
```
Baggerone, line 878: Did you mean for this to be `getLedgerEntryType`

Baggerone (approved):

application/grifts/import.go
Schparky: I think it predates the creation of `type LedgerEntryRecordType string`. It should be updated.

PR #164, opened 2021-10-04
Get and convert recently updated items

Baggerone:
Partial progress towards new endpoints


application/models/testutils.go
```diff
@@ -503,3 +503,20 @@ func RegisterEventDetector(kind string, detected *bool) (events.DeleteFn, error)
 		}
 	})
 }
+
+func assertMapsUUIDTimeEqual(want, got map[uuid.UUID]time.Time) string {
```
Schparky, line 507: Looks like I could have removed this as well.

Schparky (approved):

PR #165, opened 2021-10-04
use a SQL query to do the grouping and sorting

Schparky:


application/models/policyhistory_test.go
```diff
@@ -67,7 +66,7 @@ func CreatePolicyHistoryFixtures_RecentItemStatusChanges(tx *pop.Connection) Fix
 	hydratePHsForItem(4, mixedNewItem.ID)
 	hydratePHsForItem(8, noneNewItem.ID)
 
-	for i, _ := range pHistories {
+	for i := range pHistories {
```
Schparky, line 69: I think this was a goFumpt edit. I don't recall making this change.

application/models/item_test.go
```diff
@@ -758,7 +758,7 @@ func (ms *ModelSuite) Test_ItemsWithRecentStatusChanges() {
 	gotRaw, gotErr := ItemsWithRecentStatusChanges(ms.DB)
 	ms.NoError(gotErr)
 
-	const tmFmt = "Jan _2 15:04:05.0000"
+	const tmFmt = "Jan _2 15:04:05.00"
```
Schparky, line 761: Not sure why the fourth decimal place differed.

application/models/policyhistory_test.go
```diff
@@ -101,45 +100,15 @@ func (ms *ModelSuite) TestPolicyHistories_RecentItemStatusChanges() {
 	var gotPHs PolicyHistories
 
 	ms.NoError(gotPHs.RecentItemStatusChanges(ms.DB), "error calling function")
-	got := make([][2]string, len(gotPHs))
+	got := make([]string, len(gotPHs))
```
Schparky, line 103: Since the raw SQL doesn't return the ID of the PolicyHistory row, this part of the test didn't work.

application/models/policyhistory_test.go
```diff
@@ -101,45 +100,15 @@ func (ms *ModelSuite) TestPolicyHistories_RecentItemStatusChanges() {
...
-		got[i] = [2]string{g.ID.String(), g.ItemID.UUID.String()}
+		got[i] = g.ItemID.UUID.String()
 	}
 
-	want := [][2]string{
-		{phFixes[2].ID.String(), phFixes[2].ItemID.UUID.String()},
-		{phFixes[3].ID.String(), phFixes[3].ItemID.UUID.String()},
-		{phFixes[7].ID.String(), phFixes[7].ItemID.UUID.String()},
+	want := []string{
+		phFixes[3].ItemID.UUID.String(),
```
Schparky, line 109: The first two have the same Item ID so the raw query only returned one of them.

PR #168, opened 2021-10-05
API endpoint for admin/recent objects

Baggerone:
I tried using our authz middleware with `/policies/recent`, but it expects there to be an ID in that second spot.
Since, it's only an endpoint for admins, I left it the way the batches group is done.

I also just moved the fixture creation functions to models.testutils.


application/actions/admin.go
```diff
@@ -0,0 +1,51 @@
...
+import (
+	"fmt"
+
+	"github.com/gobuffalo/buffalo"
+
+	"github.com/silinternational/cover-api/api"
+	"github.com/silinternational/cover-api/models"
+)
+
+// swagger:operation GET /admin/recent Admin ListRecentObjects
```
Schparky, line 12: I think the UI is using "steward" for all their paths (routes) not intended for customers. This is probably in line with Alex's comments about "admin" bearing connotations of IT terminology. Since this is not user-facing it may not be a problem to leave this as "admin". However, it might be better to stay in sync when possible.

application/actions/admin.go
```diff
@@ -0,0 +1,51 @@
...
+	if err != nil {
+		return reportError(c, err)
+	}
+
+	claims, err := models.ClaimsWithRecentStatusChanges(tx)
+	if err != nil {
+		return reportError(c, err)
+	}
+
+	recent := api.RecentObjects{
```
Schparky, line 45: Nice and clean. Looks good.

application/models/testutils.go
```diff
@@ -528,3 +528,191 @@ func RegisterEventDetector(kind string, detected *bool) (events.DeleteFn, error)
...
 	})
 }
+
+// CreatePolicyHistoryFixtures generates a Policy with three Items each with
+//   four PolicyHistory entries as follows
+//	 CoverageStatus/Create  [not included because not update]
+//	 Name/Update [not included because not on CoverageStatus field]
+//	 CoverageStatus/Update [could be included, if date is recent]
+//	 CoverageStatus/Update [could be included, if date is recent]
+func CreatePolicyHistoryFixtures_RecentItemStatusChanges(tx *pop.Connection) Fixtures {
```
Schparky, line 538: Just thinking out loud here: it's good to have this here so other packages can use it. However, I can see this file becoming exceedingly large. I wonder how we could organize it.

Schparky (approved):

application/actions/admin.go
Baggerone: :+1:

application/models/testutils.go
Baggerone: I used to worry about that type of issue, but I'm not sure that it affects me on a day to day  basis.
Our IDEs' help us find what we need pretty quickly.

I guess we could have a file for Policy & Item fixtures and a different one for Claim fixtures and a third one for other types of fixtures.

Anyway, maybe we can leave that for another time.

Schparky: Definitely can wait for another time.

jason-jackson (approved):

PR #174, opened 2021-10-07
compress migrations down to table create migration files

Schparky:


Baggerone (approved):
Semi-Easy-Button

application/migrations/20210803150753_create_item_categories.up.fizz
```diff
@@ -13,3 +14,9 @@ create_table("item_categories") {
 
 	t.ForeignKey("risk_category_id", {"risk_categories": ["id"]}, {"on_delete": "cascade"})
 }
+
+sql(`
```
jason-jackson, line 18: I don't see any inserts into item_categories before calling this. Will this do anything?

Schparky: No, I suppose not. When that was added, it assumed there were records already in the database. As an initial migration it no longer makes sense. Thanks for identifying that.

PR #193, opened 2021-10-15
stubs for premium charge calculation

Schparky:
taking this in bite-size pieces


application/models/item.go
```diff
@@ -81,15 +81,15 @@ func (i *Item) Update(ctx context.Context) error {
 	if err := oldItem.FindByID(tx, i.ID); err != nil {
 		return appErrorFromDB(err, api.ErrorQueryFailure)
 	}
-	validTrans, err := isItemTransitionValid(oldItem.CoverageStatus, i.CoverageStatus)
-	if err != nil {
+	if validTrans, err := isItemTransitionValid(oldItem.CoverageStatus, i.CoverageStatus); err != nil {
```
Baggerone, line 84: Does this do anything differently than what was there?

PR #230, opened 2021-11-01
filter policies by whether or not they have approved items

Schparky:
This PR does not change the default behavior, though we may want to consider a default of `search=active:true` for admins.

### DIFF
--- a/application/models/policy.go
+++ b/application/models/policy.go
@@ -297,9 +297,8 @@ func scopeSearchPoliciesByActive(active string) pop.ScopeFunc {
 				"WHERE policies.id=items.policy_id AND items.coverage_status='Approved')")
 		}
 		if active == "false" {
-			return q.Where("policies.id IN (SELECT policies.id FROM policies LEFT JOIN items " +
-				"ON policies.id=items.policy_id " +
-				"WHERE items.coverage_status!='Approved' AND items.id IS NULL)")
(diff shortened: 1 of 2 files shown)


application/models/policy.go
```diff
@@ -286,6 +290,21 @@ func scopeSearchPolicyByName(name string) pop.ScopeFunc {
...
+func scopeSearchPoliciesByActive(active string) pop.ScopeFunc {
+	return func(q *pop.Query) *pop.Query {
+		if active == "true" {
+			return q.Where("policies.id IN (SELECT policies.id FROM policies,items " +
+				"WHERE policies.id=items.policy_id AND items.coverage_status='Approved')")
+		}
+		if active == "false" {
+			return q.Where("policies.id IN (SELECT policies.id FROM policies LEFT JOIN items " +
+				"ON policies.id=items.policy_id " +
+				"WHERE items.coverage_status!='Approved' AND items.id IS NULL)")
```
Baggerone, line 302: I don't understand what the `AND items.id IS NULL` is there for.  Is that what you intended?

Baggerone (approved):
Approved but with a potential change to make first, unless I'm just confused.

application/models/policy.go
Schparky: Good question. I thought I had it working as a subtraction of the set of records that have no matching items, but it was a broken attempt at that. I think it's fixed now.

PR #237, opened 2021-11-03
Feature/delete dependent

Baggerone:


application/actions/dependents.go
```diff
@@ -124,6 +126,41 @@ func dependentsUpdate(c buffalo.Context) error {
...
+//     description: policy dependent ID
+// responses:
+//   '204':
+//     description: OK but no content in response
+func dependentsDelete(c buffalo.Context) error {
+	tx := models.Tx(c)
+	dependent := getReferencedDependentFromCtx(c)
+
+	relatedItemNames := dependent.RelatedItemNames(tx)
+	if len(relatedItemNames) == 0 {
```
Schparky, line 149: In my opinion, it would read better to handle the error first.
```suggestion
	if len(relatedItemNames) >  0 {
```

application/models/policydependent.go
```diff
@@ -55,6 +56,24 @@ func (p *PolicyDependent) Update(tx *pop.Connection) error {
 	return update(tx, p)
 }
 
+func (p *PolicyDependent) Destroy(tx *pop.Connection) error {
+	return destroy(tx, p)
+}
+
+// canBeDeleted checks for related items
```
Schparky, line 63:
```suggestion
// RelatedItemNames checks for related items
```

Schparky (approved):

application/models/policydependent.go
```diff
@@ -55,6 +56,24 @@ func (p *PolicyDependent) Update(tx *pop.Connection) error {
...
+func (p *PolicyDependent) RelatedItemNames(tx *pop.Connection) []string {
```
jason-jackson, line 64: I think it would be better to use RelatedItems() rather than just names, as we'll want more than just names in emails.

```suggestion
func (p *PolicyDependent) RelatedItems(tx *pop.Connection) []Items {
```

jason-jackson (changes requested):

jason-jackson (approved):

PR #238, opened 2021-11-04
email template fixes

Schparky:
By observation, it appears that gobuffalo/plush cannot handle custom string types (e.g. `type myType string`) even if they have a `String()` function. On the flip side, `nulls.String` seems to work, so it's not exactly clear where the difference lies. Also, function calls like `item.GetMakeModel()` appear to work, so we may want to move some of these one-off template variables into model functions (e.g. `item.AnnualPremiumForTemplate()`)

### DIFF
--- a/application/messages/claim.go
+++ b/application/messages/claim.go
@@ -2,13 +2,11 @@ package messages
 
 import (
 	"fmt"
-	"time"
 
 	"github.com/gobuffalo/nulls"
 	"github.com/gobuffalo/pop/v5"
 
 	"github.com/silinternational/cover-api/api"
-	"github.com/silinternational/cover-api/domain"
 	"github.com/silinternational/cover-api/models"
 )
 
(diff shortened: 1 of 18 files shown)


application/messages/claim.go
```diff
@@ -19,21 +17,13 @@ func ClaimReview1QueueMessage(tx *pop.Connection, claim models.Claim) {
...
-	data["item"] = claim.ClaimItems[0].Item
-
-	data["payoutOptionDescription"] = api.PayoutOptionDescriptions[claim.ClaimItems[0].PayoutOption]
-	data["maximumPayout"] = 0 // TODO: calculate this value
-	data["submitted"] = domain.TimeBetween(time.Now().UTC(), claim.UpdatedAt)
-
 	notn := models.Notification{
 		ClaimID: nulls.NewUUID(claim.ID),
 		Body:    data.renderHTML(MessageTemplateClaimReview1Steward),
-		Subject: "Action Required. " + memberName + " just (re)submitted a claim for approval",
```
Baggerone, line 36: Did someone say to remove the `(re)`?  This will be hit the first time they submit the claim but also after the claim is in `revision` status and they resubmit it, right?

application/messages/messages.go
```diff
@@ -74,14 +85,21 @@ func (m MessageData) addItemData(tx *pop.Connection, item models.Item) {
 	m["supportEmail"] = steward.Email
 	m["supportName"] = steward.Name()
 
-	m["itemURL"] = fmt.Sprintf("%s/items/%s", domain.Env.UIURL, item.ID)
+	m["itemURL"] = fmt.Sprintf("%s/policies/%s/items/%s", domain.Env.UIURL, item.PolicyID, item.ID)
```
fillup, line 88: Does this work with the AuthZ middleware? I don't recall it supporting the item id in this case.

fillup (approved):

Baggerone (approved):

application/messages/messages.go
Schparky: Commenting here for the record, it's a UI URL, so that doesn't apply.

application/messages/claim.go
Schparky: It was my call. I figured it will happen much more as a first time, and the second time still reads OK in my opinion.

PR #241, opened 2021-11-04
assorted fixes

Schparky:


application/grifts/db.go
```diff
@@ -409,9 +418,9 @@ func createClaimFixtures(tx *pop.Connection, fixPolicies []*models.Policy, items
 		}
 
 		ci := models.ClaimItem{
-			ID:           domain.GetUUID(),
+			ID:           uuid.FromStringOrNil(claimItemUUIDs[i]),
 			ClaimID:      fixClaims[i].ID,
-			ItemID:       items[i].ID,
+			ItemID:       items[i*2].ID,
```
Baggerone, line 423: Does the `if` block in line 396 need to be modified to match this doubling?

Baggerone (approved):

application/grifts/db.go
Schparky: looks like it. thanks

PR #242, opened 2021-11-04
return accountable person as an object (not just an ID)

Schparky:


Baggerone (approved):

application/models/item.go
```diff
@@ -795,40 +801,40 @@ func (i *Item) CreateLedgerEntry(tx *pop.Connection, entryType LedgerEntryType,
...
 	}
-	return person.GetLocation(), nil
+	return person.GetLocation()
 }
 
 // GetAccountablePerson gets the accountable person as a Person interface
-func (i *Item) GetAccountablePerson(tx *pop.Connection) (Person, error) {
+func (i *Item) GetAccountablePerson(tx *pop.Connection) Person {
 	var person Person
 	if i.PolicyUserID.Valid {
```
jason-jackson, line 823: You didn't modify this, just noting this now.

This could be simplified down to just one sql call by switching statements around (if wanting to preserve results, and using either a return or else.

jason-jackson (approved):

application/models/item.go
Schparky: Only one of the two should ever be valid, so it would be very rare to make both database queries. Or am I missing what you're saying?

PR #248, opened 2021-11-09
don't allow an update on an item with an active claim

Schparky:


application/models/item.go
```diff
@@ -841,6 +848,29 @@ func (i *Item) GetMakeModel() string {
...
+		api.ClaimItemStatusDraft,
+		api.ClaimItemStatusPaid,
+		api.ClaimItemStatusDenied,
+	}
+
+	var claimItems ClaimItems
+	n, err := tx.Where("item_id = ?", i.ID).
+		Where("status NOT IN (?)", SafeClaimItemStatuses).
+		Count(&claimItems)
+	if err != nil {
```
Baggerone, line 865: Just wondering if you would want a panic here and avoid returning an error/nil value.

Baggerone (approved):

application/models/item.go
Schparky: yep, thanks

PR #254, opened 2021-11-10
Coerce dependent relationship on a team-type policy

Baggerone:


application/actions/dependents_test.go
```diff
@@ -320,9 +357,15 @@ func (as *ActionSuite) Test_DependentsUpdate() {
 			var dependent models.PolicyDependent
 			as.NoError(as.DB.Find(&dependent, tt.oldDep.ID), "error finding newly updated dependent.")
 			as.Equal(tt.input.Name, dependent.Name, "incorrect Name")
-			as.Equal(tt.input.Relationship, dependent.Relationship, "incorrect Relationship")
 			as.Equal(tt.input.Country, dependent.Country, "incorrect Country")
-			as.Equal(tt.input.ChildBirthYear, dependent.ChildBirthYear, "incorrect ChildBirthYear")
+
+			if tt.wantRelationship == "" {
```
Schparky, line 362: I was confused about this for a bit. Would it make more sense to put the actual expected value of the relationship in the `tt.wantRelationship` so you can always test against it rather than doing this if/else?

fillup (approved):

application/models/testutils.go
```diff
@@ -390,6 +390,58 @@ func CreatePolicyFixtures(tx *pop.Connection, config FixturesConfig) Fixtures {
 	}
 }
 
+// CreateTeamPolicyFixtures generates any number of policy records and associated policy users
```
Schparky, line 393: Rather that create a new function that is a near-duplicate of an existing one (yet another place to update when something changes in `models.Policy`) can you use `ConvertPolicyType`?

Schparky (changes requested):

Schparky:
Would it be possible to create a "None" dependent on a HH policy? An alternate approach would be to change the value of `api.PolicyDependentRelationshipNone` to an empty string. Then add validation checks to ensure the relationships are valid either way.

Baggerone:
The problem with trying to validate the relationship on a HH policy, is that the validator doesn't know what type of policy it is.

Do we want to add extra validation in the model itself for that?

application/models/testutils.go
Baggerone: Oops, I forgot about that function.

Schparky:
> The problem with trying to validate the relationship on a HH policy, is that the validator doesn't know what type of policy it is.
> 
> Do we want to add extra validation in the model itself for that?

We validate ClaimItems based on Claim fields, so I think it would work here too. I was a little dubious, so I checked it out and it seems to work. Maybe because we always hydrate a ClaimItem before passing it to a `Create` or `Update` function. Or maybe Pop does that automatically. Not sure.

Schparky (approved):

PR #258, opened 2021-11-11
todo cleanup

Schparky:


application/domain/domain.go
```diff
@@ -213,6 +215,7 @@ func readEnv() {
 	Env.PolicyMaxCoverage *= CurrencyFactor
 	Env.DependentAutoApproveMax *= CurrencyFactor
 	Env.PremiumMinimum *= CurrencyFactor
+	Env.DeductibleString = fmt.Sprintf("%.1f%%", Env.Deductible*100)
```
mtompset, line 218: Is 1 decimal place enough? It can't be 5.25% with only 1 decimal place. And quarters seems like a reasonable thing to have.

application/models/claimitem.go
```diff
@@ -399,7 +399,7 @@ func (c *ClaimItem) updatePayoutAmount(ctx context.Context) error {
 
 	coverageAmount := c.Item.CoverageAmount
 
-	deductible := 0.05
+	deductible := domain.Env.Deductible
```
mtompset, line 402: I like the move to a configurable deductible. :)

Baggerone (approved):

mtompset (commented):
Configurable deductible: nice. Single decimal? Why not 2 for 5.25%? And who doesn't like removing TODO's. :)

application/domain/domain.go
Schparky: You're right. However, that means that 5% would display as 5.00% which I don't like. And I don't particularly care to write code to round to a "sensible" number of digits.

mtompset: %6g%% might do the trick.

PR #259, opened 2021-11-11
Check that a jump from review1 to review3 has the right payout option

Baggerone:


application/actions/claims_test.go
```diff
@@ -946,7 +946,16 @@ func (as *ActionSuite) Test_ClaimsApprove() {
...
 	draftClaim := policy.Claims[0]
-	review1Claim := models.UpdateClaimStatus(as.DB, policy.Claims[1], api.ClaimStatusReview1, "")
+
+	// Make one of the claims requesting a FixedFraction payout
+	ffClaim := models.UpdateClaimStatus(as.DB, policy.Claims[1], api.ClaimStatusReview1, "")
+	ffClaim.IncidentType = api.ClaimIncidentTypeEvacuation
+	as.NoError(as.DB.Update(&ffClaim), "error updating claim fixture")
+	ffClaim.LoadClaimItems(as.DB, false)
+	ffItem := ffClaim.ClaimItems[0]
+	ffItem.PayoutOption = api.PayoutOptionFixedFraction
```
Schparky, line 956: `UpdateClaimItems` is available as a shortcut for this

application/actions/claims_test.go
```diff
@@ -968,17 +977,17 @@ func (as *ActionSuite) Test_ClaimsApprove() {
...
 			actor:      policyCreator,
-			oldClaim:   review1Claim,
+			oldClaim:   ffClaim,
 			wantStatus: http.StatusNotFound,
 		},
 		{
 			name:            "review1 to review3",
 			actor:           appAdmin,
-			oldClaim:        review1Claim,
+			oldClaim:        ffClaim,
```
Schparky, line 986: Maybe somewhat irrelevant since it's already here, but this may be a redundant test with what is done in the `models` package.

Schparky (approved):

application/actions/claims_test.go
Baggerone: Well, I used a different payout option here.

PR #265, opened 2021-11-12
fix Claim.SubmittedAt algorithm

Baggerone:


application/models/claim.go
```diff
@@ -895,9 +895,30 @@ func (c *Claim) UpdateStatus(ctx context.Context, newStatus api.ClaimStatus) err
...
+//   ClaimHistory for this claim that is for a Status Update where tne
+//   new field is Review1.  Uses the CreatedAt time of the earliest history
+//   with Status change to Review1.
 func (c *Claim) SubmittedAt(tx *pop.Connection) time.Time {
-	// TODO: use the history table to get the real date
-	return c.UpdatedAt
+	var histories ClaimHistories
+
+	err := tx.RawQuery(`
+SELECT created_at
```
Schparky, line 906: If you're concerned about extraneous whitespace, you can do this:
```suggestion
	err := tx.RawQuery("SELECT created_at"+
		" FROM claim_histories"+
		" WHERE claim_id = ? AND field_name = ? AND action = ? AND new_value = ?"+
		" ORDER BY created_at ASC",
```
otherwise, just indent each line a couple places to line up with the surrounding code.

application/models/claim.go
```diff
@@ -895,9 +895,30 @@ func (c *Claim) UpdateStatus(ctx context.Context, newStatus api.ClaimStatus) err
...
+FROM claim_histories
+WHERE claim_id = ? AND field_name = ? AND action = ? AND new_value = ?
+ORDER BY created_at ASC
+`, c.ID, FieldClaimStatus, api.HistoryActionUpdate, api.ClaimStatusReview1).All(&histories)
```
jason-jackson, line 910: Is it necessary to get all, wouldn't just first one suffice?

Schparky (approved):

application/models/claim.go
Baggerone: :+1:

application/models/claim.go
Baggerone: :+1:

PR #267, opened 2021-11-12
Change refund granularity to monthly

Baggerone:
It was easy enough to distinguish between refunds for old items stopped in January and newly covered items stopped in January, so I implemented that.  Old items stopped in January get full refund.  New items stopped in January get 11/12 refund, similar to the other months.

### DIFF
--- a/application/domain/domain.go
+++ b/application/domain/domain.go
@@ -422,6 +422,15 @@ func CalculatePartialYearValue(value int, startDate time.Time) int {
 	return int(math.Round(float64(value*daysSince) / float64(days)))
 }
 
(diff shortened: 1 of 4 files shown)


Schparky (commented):
You make this look easy.

Schparky (approved):

Baggerone:
> You make this look easy.

Heh, heh. I'm glad you said that, because I think I made it too simplistic. It looks like the calculation for premium is based on the whole year, even if they had only started part of the year in.

Schparky (approved):
Good catch. Looks like the previous implementation failed to take that into account as well.

PR #268, opened 2021-11-12
look up household using an API call

Schparky:


application/listeners/listeners.go
```diff
@@ -120,3 +122,34 @@ func findObject(payload events.Payload, object interface{}, listenerName string)
 func getDelayDuration(multiplier int) time.Duration {
 	return time.Duration(domain.Env.ListenerDelayMilliseconds) * time.Millisecond * time.Duration(multiplier)
 }
+
+func GetHHID(staffID string) string {
+	if domain.Env.HouseholdIDLookupURL == "" {
+		return ""
+	}
+
+	req, err := http.NewRequest(http.MethodGet, domain.Env.HouseholdIDLookupURL+staffID, nil)
```
Baggerone, line 131: Just wondering if this should be made resilient to a missing slash on the end of HouseholdIDLookupURL

application/listeners/listeners.go
```diff
@@ -120,3 +122,34 @@ func findObject(payload events.Payload, object interface{}, listenerName string)
...
+	if err != nil {
+		domain.ErrLogger.Printf("HHID API error, %s", err)
+		return ""
+	}
+	req.SetBasicAuth(domain.Env.HouseholdIDLookupUsername, domain.Env.HouseholdIDLookupPassword)
+
+	client := &http.Client{Timeout: time.Second * 10}
```
Baggerone, line 138: Is 10 seconds long enough?

application/listeners/listeners_test.go
```diff
@@ -114,3 +114,30 @@ func (ts *TestSuite) Test_findObject() {
 		})
 	}
 }
+
+func (ts *TestSuite) TestGetHHID() {
```
Baggerone, line 118: Would it be better to have a Skip statement in an `if` block at the start of the Test, instead of effectively skipping it silently down inside the test loop?

Baggerone (approved):

application/listeners/listeners.go
Schparky: The longest I've seen is about 5 seconds. I'm not sure how to choose this timeout value.

PR #273, opened 2021-11-15
use a single household entity for all household policies

Schparky:


application/models/entitycode.go
```diff
@@ -51,3 +53,7 @@ func (ec *EntityCode) ConvertToAPI(tx *pop.Connection) api.EntityCode {
 		Name: fmt.Sprintf("%s - %s", ec.Code, ec.Name),
 	}
 }
+
+func HouseholdEntityID() uuid.UUID {
+	return uuid.FromStringOrNil(HouseholdEntityIDString)
```
Baggerone, line 58: I wonder if it would be preferable just to make this a file level `var`, instead of recreating it every time it's needed.
(You could still make the var unexported and keep this function to return it.)

Baggerone (approved):

application/models/entitycode.go
Schparky: sounds good

PR #278, opened 2021-11-16
Repeating job to inactivate certain items

Baggerone:


application/job/job.go
```diff
@@ -0,0 +1,88 @@
...
+	"github.com/silinternational/cover-api/models"
+)
+
+const (
+	InactivateItems = "inactivate_items"
+)
+
+var w worker.Worker
+
+// TestBuffaloContext is a buffalo context user in tests
```
Schparky, line 19:
```suggestion
// jobBuffaloContext is a buffalo context for jobs
```

application/job/job.go
```diff
@@ -0,0 +1,88 @@
...
+func (j *jobBuffaloContext) Value(key interface{}) interface{} {
+	return j.params[key]
+}
+
+// Set sets the value to be associated with the given key in the test context
+func (j *jobBuffaloContext) Set(key string, val interface{}) {
+	j.params[key] = val
+}
+
+// CreateTestContext sets the domain.ContextKeyCurrentUser to the user param in the TestBuffaloContext
```
Schparky, line 35:
```suggestion
// createJobContext creates an empty context
```

application/models/item.go
```diff
@@ -263,18 +263,19 @@ func (i *Item) FindByID(tx *pop.Connection, id uuid.UUID) error {
 //  and if it has a Draft, Revision or Pending status.
 //  If the item's status is Denied or Inactive, it does nothing.
 //  Otherwise, it changes its status to Inactive.
-func (i *Item) SafeDeleteOrInactivate(ctx context.Context, actor User) error {
+func (i *Item) SafeDeleteOrInactivate(ctx context.Context) error {
+	now := time.Now()
```
Schparky, line 267:
```suggestion
	now := time.Now().UTC()
```

application/models/item.go
```diff
@@ -313,20 +314,34 @@ func (i *Item) isNewEnough() bool {
 	return !i.CreatedAt.Before(cutOffDate)
 }
 
-// Inactivate sets the item's CoverageStatus to Inactive
-//  TODO deal with coverage payment changes
-func (i *Item) Inactivate(ctx context.Context) error {
+// ScheduleInactivation sets the item's CoverageEndDate and creates a corresponding
+//  credit ledger entry
+func (i *Item) ScheduleInactivation(ctx context.Context, t time.Time) error {
 	i.CoverageStatus = api.ItemCoverageStatusInactive
```
Schparky, line 320: Should the status remain unchanged until the worker runs?

application/models/item.go
```diff
@@ -313,20 +314,34 @@ func (i *Item) isNewEnough() bool {
...
 
 	user := CurrentUser(ctx)
 	i.StatusChange = ItemStatusChangeInactivated + user.Name()
+
+	// If now it's January and the item was created before this year
+	//   set its CoverageEndDate to the current day.
+	// Otherwise, set it to the end of the current month.
+	if i.CoverageStartDate.Year() < t.Year() && t.Month() == 1 {
```
Schparky, line 328: It looks like this logic is tied to the decision to allow complete refunds in January (i.e. not bill for the remainder of the month.) To couple this decision with the premium refund calculation, maybe a `const` or `Env` value would be helpful, e.g.: `const AllowJanuaryPremiumRefund = true`. Then reference that in both places.

application/models/item.go
```diff
@@ -263,18 +263,19 @@ func (i *Item) FindByID(tx *pop.Connection, id uuid.UUID) error {
...
 //  Otherwise, it changes its status to Inactive.
-func (i *Item) SafeDeleteOrInactivate(ctx context.Context, actor User) error {
+func (i *Item) SafeDeleteOrInactivate(ctx context.Context) error {
+	now := time.Now()
 	switch i.CoverageStatus {
 	case api.ItemCoverageStatusInactive, api.ItemCoverageStatusDenied:
 		return nil
 	case api.ItemCoverageStatusApproved:
-		return i.Inactivate(ctx)
+		return i.ScheduleInactivation(ctx, now)
```
Schparky, line 272: What is the advantage to scheduling inactivation for "now" versus just calling `Inactivate`?

Schparky (approved):

application/models/item.go
Baggerone: This was probably just a name change hold-over.  I'll switch it to the new Inactivate function.

Baggerone: Oops, actually the scheduling function also does the credit ledger entry.

application/models/item.go
Baggerone: Good catch. Missed that one.

application/models/item.go
Schparky: Sounds like that ought to be clarified by refactoring or changing function name(s).

Schparky (approved):

application/models/item.go
```diff
@@ -683,30 +725,42 @@ func (i *Item) calculateProratedPremium(t time.Time) api.Currency {
 	return api.Currency(p)
 }
 
+// True if coverage on the item started in a previous year and the current
+//  month is January.
+func (i *Item) giveFullYearRefund(t time.Time) bool {
```
fillup, line 730: I think this method should be renamed to something like `shouldGiveFullYearRefund` so it's more of a question rather than a command.

application/models/item.go
```diff
@@ -663,6 +676,35 @@ func (i *Item) ConvertToAPI(tx *pop.Connection) api.Item {
 	return apiItem
 }
 
+// InactivateApprovedButEnded fetches all the items that have coverage_status=Approved
+//   and coverage_end_date before today and then
+//   saves them with coverage_status=Inactive
+func (i *Items) InactivateApprovedButEnded(ctx context.Context) error {
+	tx := Tx(ctx)
+
+	endDate := time.Now().Format(domain.DateFormat)
```
fillup, line 685: Do we really have to format the date before passing into query for comparison?

fillup: Also, make it `time.Now().UTC().Format...`

application/models/item.go
```diff
@@ -683,30 +725,42 @@ func (i *Item) calculateProratedPremium(t time.Time) api.Currency {
...
+	var premium int
+
 	// If the coverage was from a previous year and today is still in January,
 	//   give a full year's refund.
-	// If today is after January, give a refund for the following months only.
+	if i.giveFullYearRefund(t) {
+		premium = int(i.CalculateAnnualPremium())
+		return api.Currency(-1 * premium)
+
+	}
```
fillup, line 748:
```suggestion
		return api.Currency(-1 * premium)
	}
```

fillup (approved):
a couple minor comments, but otherwise looks good

application/models/item.go
Baggerone: I'm thinking that truncating it to just the date will avoid things from earlier "today" being included, rather than leaving those for tomorrow's run.

Is it better the other way?

application/models/item.go
Baggerone: :+1:

PR #281, opened 2021-11-17
avoid using nulls types in api package

Schparky:
goswagger tries to use the docs in the nulls package as documentation for our API. This should avoid that.

### DIFF
--- a/application/models/entitycode.go
+++ b/application/models/entitycode.go
@@ -12,7 +12,7 @@ import (
 
 const HouseholdEntityIDString = "5f181e39-0a2a-49ac-8796-2f3a3de9fcbd"
 
-var HouseholdEntityID uuid.UUID
+var householdEntityID uuid.UUID
 
 type EntityCodes []EntityCode
 
@@ -55,4 +55,8 @@ func (ec *EntityCode) ConvertToAPI(tx *pop.Connection) api.EntityCode {
 		Code: ec.Code,
 		Name: fmt.Sprintf("%s - %s", ec.Code, ec.Name),
 	}
+}
+
+func HouseholdEntityID() uuid.UUID {
+	return householdEntityID
 }
--- a/application/models/entitycode.go
+++ b/application/models/entitycode.go
@@ -12,7 +12,7 @@ import (
 
 const HouseholdEntityIDString = "5f181e39-0a2a-49ac-8796-2f3a3de9fcbd"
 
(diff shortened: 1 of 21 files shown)


application/models/riskcategory.go
```diff
@@ -16,8 +16,8 @@ const (
 )
 
 var (
-	riskCategoryMobileID     = uuid.FromStringOrNil(RiskCategoryMobileIDString)
-	riskCategoryStationaryID = uuid.FromStringOrNil(RiskCategoryStationaryIDString)
+	RiskCategoryMobileID     = uuid.FromStringOrNil(RiskCategoryMobileIDString)
```
Baggerone, line 19: Isn't it better to leave these unexported, and keep getter the functions, in order to ensure other packages don't modify the values?

application/models/models.go
```diff
@@ -187,9 +188,9 @@ func init() {
 	mValidate.RegisterStructValidation(notificationStructLevelValidation, Notification{})
 
 	// get fixed IDs
-	riskCategoryStationaryID = uuid.FromStringOrNil(RiskCategoryStationaryIDString)
-	riskCategoryMobileID = uuid.FromStringOrNil(RiskCategoryMobileIDString)
-	householdEntityID = uuid.FromStringOrNil(HouseholdEntityIDString)
+	RiskCategoryStationaryID = uuid.FromStringOrNil(RiskCategoryStationaryIDString)
```
Baggerone, line 191: Isn't it better to leave these unexported, and keep getter the functions, in order to ensure other packages don't modify the values?

application/models/entitycode.go
```diff
@@ -12,7 +12,7 @@ import (
 
 const HouseholdEntityIDString = "5f181e39-0a2a-49ac-8796-2f3a3de9fcbd"
 
-var householdEntityID uuid.UUID
+var HouseholdEntityID uuid.UUID
```
Baggerone, line 15: Isn't it better to leave these unexported, and keep getter the functions, in order to ensure other packages don't modify the values?

Baggerone (approved):
I copied and pasted the same question in a few places.

application/models/riskcategory.go
Schparky: Good point. I was just going for simplicity since I had to get a pointer to one of them. Guess it's worth the extra complexity.

PR #284, opened 2021-11-17
item.coverage_paid_through

Baggerone:


application/models/item.go
```diff
@@ -577,7 +590,12 @@ func (i *Item) Approve(ctx context.Context, doEmitEvent bool) error {
 	}
 
 	amount := i.calculateProratedPremium(time.Now().UTC())
-	return i.CreateLedgerEntry(Tx(ctx), LedgerEntryTypeNewCoverage, amount)
+	if err := i.CreateLedgerEntry(Tx(ctx), LedgerEntryTypeNewCoverage, amount); err != nil {
+		return err
+	}
+
+	i.CoveragePaidThrough = nulls.NewTime(domain.EndOfYear())
```
Schparky, line 597: To be safe, this should probably be the end of the same year that is used inside `CreateLedgerEntry` as the `DateSubmitted`. (Side note: I think there's a bug in there. It uses `CoverageStartDate` even for coverage changes.) To do that, and to be less fragile to future changes, `CoveragePaidThrough` should probably be set in `CreateLedgerEntry`.

application/models/item.go
```diff
@@ -276,7 +279,17 @@ func (i *Item) SafeDeleteOrInactivate(ctx context.Context) error {
 		if err := i.ScheduleInactivation(ctx, now); err != nil {
 			return err
 		}
-		return i.CreateLedgerEntry(Tx(ctx), LedgerEntryTypeCoverageChange, i.calculateCancellationCredit(now))
+
+		// Don't give a refund if coverage has not been paid through the end of the year
+		if !i.CoveragePaidThrough.Valid || i.CoveragePaidThrough.Time.Year() < time.Now().UTC().Year() {
```
Schparky, line 284: Since this only works if `CoveragePaidThrough` is set to the end of the year, I wonder if that field should just be an integer rather than a date. Something like `PaidThroughYear`. ("Paid" implies coverage since that's the only kind of payment.) Also, it may not need to be null since a default of 0 would be handled correctly by `if i.PaidThroughYear < time.Now().UTC().Year()`

Schparky (approved):
I think this will work as-is, but I made some suggestions that might make it a tad more safe for corner cases and timing issues.

Schparky:
Another thought. Maybe while this is fresh, make a `RenewCoverage` function that bumps the coverage/premium paid field and adds another ledger entry.

application/models/item.go
Baggerone: If we move the setting/clearing of  `PaidThroughYear` to the CreateLedgerEntry() function, which would you prefer ...?

- either set it or clear it based on whether the ledger amount is positive or negative
- add a param to tell it to clear out the `PaidThroughYear`  value

application/models/item.go
```diff
@@ -281,14 +281,14 @@ func (i *Item) SafeDeleteOrInactivate(ctx context.Context) error {
...
-		if !i.CoveragePaidThrough.Valid || i.CoveragePaidThrough.Time.Year() < time.Now().UTC().Year() {
+		if i.PaidThroughYear < time.Now().UTC().Year() {
 			return nil
 		}
 
 		if err := i.CreateLedgerEntry(Tx(ctx), LedgerEntryTypeCoverageChange, i.calculateCancellationCredit(now)); err != nil {
 			return err
 		}
-		i.CoveragePaidThrough = nulls.Time{}
+		i.PaidThroughYear = 0
```
Schparky: I don't know if it makes a difference yet, but should this be decremented instead of zeroed? Might need to be conditional depending on when coverage started.

application/models/item.go
```diff
@@ -594,7 +594,7 @@ func (i *Item) Approve(ctx context.Context, doEmitEvent bool) error {
 		return err
 	}
 
-	i.CoveragePaidThrough = nulls.NewTime(domain.EndOfYear())
+	i.PaidThroughYear = time.Now().UTC().Year()
```
Schparky: It would be an extremely unlikely event that this would be a different year than the premium calculation, but on principal I might suggest to use the same time that was captured a few lines up.

Schparky (approved):
Looks good; I just added a couple little thoughts.

application/models/item.go
Baggerone: I think I'll leave that for another day.  If what they had paid in the past only covered through a past date, then currently they don't have any kind of credit with us.

application/models/item.go
Baggerone: (Looks like Github things this is on line 597 but for me it's on line 897)

I'm not sure what time you're referring to.  The i.CoverageStartDate may be from previous years, right? Was that the bug you mentioned before? What should le.DateSubmitted based on?

Baggerone: Since NewLedgerEntry sets the DateSubmitted to the current time, maybe 
`	le.DateSubmitted = i.CoverageStartDate`
should be deleted?

PR #285, opened 2021-11-18
fix bug: item ledger entry date submitted was set to CoverageStartDate 

Schparky:


application/models/item.go
```diff
@@ -879,22 +878,23 @@ func (i *Item) CreateLedgerEntry(tx *pop.Connection, entryType LedgerEntryType,
...
+	now := time.Now().UTC()
 	name := i.GetAccountablePersonName(tx)
 
 	le := NewLedgerEntry(i.Policy, i, nil)
 	le.Type = entryType
 	le.Amount = amount
 	le.FirstName = name.First
 	le.LastName = name.Last
-	le.DateSubmitted = i.CoverageStartDate
+	le.DateSubmitted = now
```
Baggerone, line 889: Is this line needed at all, since it is done in `NewLedgerEntry`?

application/models/item.go
```diff
@@ -879,22 +878,23 @@ func (i *Item) CreateLedgerEntry(tx *pop.Connection, entryType LedgerEntryType,
...
 
 	if err := le.Create(tx); err != nil {
 		return err
 	}
 
 	oldPaidYear := i.PaidThroughYear
 	if le.Type == LedgerEntryTypeNewCoverage {
-		i.PaidThroughYear = time.Now().UTC().Year()
+		i.PaidThroughYear = now.Year()
```
Baggerone, line 897: I think if you remove the `le.DateSubmitted = now` line, you can make this `i.PaidThroughYear = le.DateSubmitted.Year()` and get rid of the line that says `now := ...`

Baggerone (approved):

PR #286, opened 2021-11-18
import items.paid_through_year

Schparky:
image


Baggerone (approved):
Approved, assuming you want to use the current year for entries where there is a null in the CoverageEndDate.  (You know more about that than me.)

PR #292, opened 2021-11-18
delete claim attachment

Baggerone:
I also moved the previous function and test for attaching a claim file to the new files


application/actions/app.go
```diff
@@ -55,6 +55,7 @@ const (
 	batchesPath         = "/batches"
 	claimsPath          = "/" + domain.TypeClaim
 	claimItemsPath      = "/" + domain.TypeClaimItem
+	claimFilesPath      = "/" + domain.TypeClaimFile
```
jason-jackson, line 58: Should be before claimItemsPath (if sorting alphabetically)

application/actions/claimfiles.go
```diff
@@ -0,0 +1,83 @@
...
+//   '204':
+//     description: OK but no content in response
+func claimFilesDelete(c buffalo.Context) error {
+	tx := models.Tx(c)
+	cFile := getReferencedClaimFileFromCtx(c)
+	cFile.Destroy(tx)
+	return c.Render(http.StatusNoContent, nil)
+}
+
+// getReferencedClaimFileFromCtx pulls the models.Item resource from context that was put there
```
Schparky, line 75:
```suggestion
// getReferencedClaimFileFromCtx pulls the models.ClaimFile resource from context that was put there
```

jason-jackson (approved):
Not all variables are alphabetized, but might be intentional.

application/models/claimfile.go
```diff
@@ -54,6 +57,56 @@ func (c *ClaimFile) Create(tx *pop.Connection) error {
...
+	}
+
+	var claim Claim
+	if err := claim.FindByID(tx, c.ClaimID); err != nil {
+		panic(err.Error())
+	}
+
+	claim.LoadPolicy(tx, false)
+	policy := claim.Policy
+	policy.LoadMembers(tx, false)
```
Schparky, line 99: I think `Policy.isMember` could be used here

Schparky (approved):

application/actions/claimfiles.go
Baggerone: That's funny> I corrected the old copy-paste error and left this one.

application/models/claimfile.go
Baggerone: :+1:

application/actions/app.go
Baggerone: :+1:

PR #306, opened 2021-11-30
Limit which fields PolicyItemsUpdate can change

Schparky:


application/actions/items.go
```diff
@@ -112,30 +112,41 @@ func itemsUpdate(c buffalo.Context) error {
...
-	if err != nil {
+	item.Name = input.Name
+	item.CategoryID = input.CategoryID
+	item.InStorage = input.InStorage
+	item.Country = input.Country
+	item.Description = input.Description
+	item.Make = input.Make
+	item.Model = input.Model
+	item.SerialNumber = input.SerialNumber
+	item.CoverageStatus = input.CoverageStatus
```
Baggerone, line 135: This shouldn't be included here, should it?

application/api/items.go
```diff
@@ -176,3 +176,47 @@ type ItemStatusInput struct {
...
+	Make string `json:"make"`
+
+	// model
+	Model string `json:"model"`
+
+	// serial number
+	SerialNumber string `json:"serial_number"`
+
+	// coverage status
+	CoverageStatus ItemCoverageStatus `json:"coverage_status"`
```
Baggerone, line 216: This shouldn't be included here, right?

Baggerone (changes requested):

jason-jackson (approved):

application/actions/items.go
```diff
@@ -102,7 +102,7 @@ func itemsCreate(c buffalo.Context) error {
 //     description: policy item create update object
```
Baggerone, line 102: delete `create` (?)

application/actions/items.go
Schparky: I guess I was thinking this endpoint was used to change status, but I forgot about the specialized endpoints for that purpose.

Baggerone (approved):

PR #307, opened 2021-12-01
decouple inactivation of ended items from other means of inactivation

Baggerone:
The lack of a real user for the history entry was causing it to fail.
When I tried to look through what was happening, I figured there was too much happening for this one little use case when the other common function was being called.

### DIFF
--- a/application/models/item.go
+++ b/application/models/item.go
@@ -694,7 +694,7 @@ func (i *Item) ConvertToAPI(tx *pop.Connection) api.Item {
 func (i *Item) inactivateEnded(ctx context.Context) error {
 	tx := Tx(ctx)
 
-	if safe := i.canBeUpdated(tx); !safe {
+	if !i.canBeUpdated(tx) {
 		err := errors.New("item cannot be updated because it has an active claim")
 		return api.NewAppError(err, api.ErrorItemHasActiveClaim, api.CategoryUser)
 	}
--- a/application/models/item.go
+++ b/application/models/item.go
@@ -99,7 +99,7 @@ func (i *Item) Update(ctx context.Context) error {
 		}
 	}
 
-	if safe := i.canBeUpdated(tx); !safe {
+	if !i.canBeUpdated(tx) {
 		err := errors.New("item cannot be updated because it has an active claim")
 		return api.NewAppError(err, api.ErrorItemHasActiveClaim, api.CategoryUser)
 	}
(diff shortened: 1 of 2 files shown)


application/models/item.go
```diff
@@ -688,6 +688,37 @@ func (i *Item) ConvertToAPI(tx *pop.Connection) api.Item {
 	return apiItem
 }
 
+// This function is only intended to be used for items that have been active
+// but are now scheduled to become inactive.
+// As such, any credit ledger entries should have already been created.
+func (i *Item) inactivateEnded(ctx context.Context) error {
+	tx := Tx(ctx)
+
+	if safe := i.canBeUpdated(tx); !safe {
```
Schparky, line 697:
```suggestion
	if !i.canBeUpdated(tx) {
```

application/models/item.go
```diff
@@ -688,6 +688,37 @@ func (i *Item) ConvertToAPI(tx *pop.Connection) api.Item {
...
+		})
+	history.ItemID = nulls.NewUUID(i.ID)
+	history.UserID = i.Policy.Members[0].ID
+	if err := history.Create(tx); err != nil {
+		return err
+	}
+
+	i.CoverageStatus = api.ItemCoverageStatusInactive
+
+	return update(tx, i)
```
Schparky, line 719: I was just thinking yesterday we could probably use a wrapper for `UpdateColumns` for patterns like this. (create a history record and update a single column) No need to do that now, but something to put in the backlog. I'll create a Trello card for it.

application/models/item_test.go
```diff
@@ -647,7 +647,7 @@ func (ms *ModelSuite) TestItem_InactivateApprovedButEnded() {
 	fixtures := CreateItemFixtures(ms.DB, fixConfig)
 	policy := fixtures.Policies[0]
 	policy.LoadItems(ms.DB, false)
-	user := policy.Members[0]
+	user := User{}
```
Schparky, line 650: I don't understand in what situation you'd have a policy with no members.

Schparky (approved):

application/models/item_test.go
Schparky: Oh, I think I see. It's not that the policy has no members. It's the job worker has no user.

application/models/item.go
Baggerone: :+1: (copy-paste legacy)

application/models/item_test.go
Baggerone: :+1:

PR #316, opened 2021-12-03
admin as actor was avoiding the status change

Baggerone:


Schparky:
I fear this may need to be more complicated. Are there changes that an admin can and should be able to make without causing it to be reverted?

Baggerone:
It's only called by the api endpoint and the status change only happens if it has Review1 status.

It was broken for the case when an was trying to edit his own claim.  It didn't get reset to `Draft` status and then the UI got an error trying to `submit` it.

Schparky (approved):

PR #318, opened 2021-12-03
redesign claim_denied_member email template

Schparky:
image

image


application/templates/mail/claim_denied_member.plush.html
```diff
@@ -10,21 +11,32 @@
...
+		incidentDate: incidentDate,
+		incidentType: incidentType,
+	}) %>
 
 	<%= partial("button", {
 		url: claimURL,
-		label: "Review Claim"}
-	) %>
+		label: "View Claim in " + appName
+	}) %>
```
Baggerone, line 41: It still needs the footer, right?

Baggerone (approved):

application/templates/mail/claim_denied_member.plush.html
Schparky: I'll circle back on that

PR #320, opened 2021-12-03
redesign of policy_user_invite email template

Schparky:
image

image

### DIFF
--- a/application/templates/mail/policy_user_invite.plush.html
+++ b/application/templates/mail/policy_user_invite.plush.html
@@ -1,26 +1,32 @@
 <div>
 	<%= partial("body_header", {
-		previewText: "",
-		title: "",
+		previewText: "Dear " + inviteeName + ", " + inviterName + " has invited you to join their insurance policy " +
+			policy.Name + ", on " + appName + " by SIL.",
+		title: "Welcome",
 		subTitle: "",
 	}) %>
 
 	<p>
 		Dear <%= inviteeName %>,
 	</p>
-	<p>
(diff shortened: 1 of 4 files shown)


application/templates/mail/policy_user_invite.plush.html
```diff
@@ -1,26 +1,32 @@
...
 	</p>
-	<p>
-		You've been invited to join <%= inviterName %>'s <%= policyType %> policy with <%= appName %>,
-the personal property protection community of SIL International. As a policy member
-you'll be able to add items for coverage, manage existing items, and file claims.
-To accept their invite, click the link below and sign in with your corporate identity account.
 
+	<p>
+		<%= inviterName %> has invited you to join their insurance policy, <%= policy.Name %>, on <%= appName %>
+		by SIL. <%= emailIntro %> As a policy member you'll be able to add items for coverage, manage existing items,
```
Baggerone, line 15: I'm thinking we don't want `SIL` hard coded here, plus it should already be in the emailIntro value

Baggerone (approved):

application/templates/mail/policy_user_invite.plush.html
Schparky: Having SIL in the code and metadata isn't a problem since it's a SIL app (github.com/silinternational). At least not in a privacy sense. I think the reason it's described like that is so it matches the "intro" text. I'm not crazy about the "by SIL" part but trying not to push back on too many things.

Schparky: The intro text here isn't exactly the same as the one in the welcome email, but it's close enough to not warrant a new environment variable.

PR #328, opened 2021-12-06
Update contents for item-revision-member email

Baggerone:
image

image

### DIFF
--- a/application/templates/mail/_item_card.plush.html
+++ b/application/templates/mail/_item_card.plush.html
@@ -17,64 +17,67 @@
 	</div>
 </div>
 
-
-<%= if ( item.SerialNumber != "" ) { %>
- <div>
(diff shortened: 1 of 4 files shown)


application/templates/mail/_item_card.plush.html
```diff
@@ -17,64 +17,67 @@
...
+		url: itemURL,
+		label: buttonLabel}
+	) %>
 
-<%= partial("button", {
-	url: itemURL,
-	label: buttonLabel}
-) %>
+	<table style="padding: 0 0 1em 0;">
+		<%= if (coverageAmount != "") { %>
```
Schparky, line 47: Is it possible for coverageAmount to be an empty string? If not, you can probably get rid of the `if`

Schparky (approved):

PR #330, opened 2021-12-06
remove email notification for item autoapproval

Schparky:
In the list of email templates, this one was tagged by the designers as unnecessary.


Schparky:
We should keep this, and possibly create a notifications preferences table to allow users to silence them.

PR #332, opened 2021-12-06
combine migration files into the create_* files

Schparky:
also added an 'import' target to the Makefiel


Baggerone (commented):
This seems like a risky thing to do right before we go live with production.
Do you have a way to compare the database structures before and after the change to ensure they're the same?

Schparky:
> This seems like a risky thing to do right before we go live with production. Do you have a way to compare the database structures before and after the change to ensure they're the same?

I compared the generated `schema.sql` file. They match, except with expected differences: order of columns, removal of defaults.

Baggerone (approved):
Approved assuming we are actually blowing away the current Prod db.

PR #333, opened 2021-12-06
set country if legacy data has a state in the country field

Schparky:


application/grifts/import.go
```diff
@@ -1149,9 +1149,19 @@ func getState(c string) string {
 func getCountry(c string) string {
 	c = trim(c)
 
-	if _, ok := states[c]; ok || len(c) <= 2 {
+	if len(c) < 2 {
 		return ""
 	}
+
+	if state := getState(c); state != "" {
+		switch c {
```
Baggerone, line 1157: Should this be `switch state`?

Schparky: probably, thanks

Baggerone (approved):

PR #343, opened 2021-12-09
Deal with related TODO's about event categories

Baggerone:
I figured that the Event values could remain as they are, since they are all different.

### DIFF
--- a/application/messages/messages.go
+++ b/application/messages/messages.go
@@ -39,6 +39,11 @@ const (
 	MessageTemplateUserWelcome      = "user_welcome"
 )
 
+const (
+	EventCategoryItem  = "Item"
+	EventCategoryClaim = "Claim"
+)
+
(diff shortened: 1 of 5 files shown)


application/messages/messages.go
```diff
@@ -39,6 +39,11 @@ const (
 	MessageTemplateUserWelcome      = "user_welcome"
 )
 
+const (
+	EventCategoryItem  = "Item"
```
Schparky, line 43: are these event categories, or message categories?

Schparky (approved):

application/messages/messages.go
Baggerone: They're the notification.EventCategory value. Which do you prefer?

Schparky: Well, I totally overlooked that. :) It's fine as it is.

PR #345, opened 2021-12-09
make claim_items.is_repairable nullable so it is not required for draft

Schparky:


Baggerone (approved):
You won't be merging this into develop until after, the dev->main PR is done, right?

Seems like this is a lot of trouble for a bool. I hope they really need it to be nullable.

Schparky:
> You won't be merging this into develop until after the dev->main PR is done, right?
> 
> Seems like this is a lot of trouble for a bool. I hope they really need it to be nullable.

I was planning on it. If you think about it, it should have been this way from the beginning. Need to have three states since Draft implies an incomplete set of field values.

PR #346, opened 2021-12-09
If a new user has a household ID that matches an existing Policy, don't create a new one for them

Baggerone:


application/models/user.go
```diff
@@ -324,6 +336,28 @@ func (u *User) CreateInitialPolicy(tx *pop.Connection, householdID string) error
 	return nil
 }
 
+func (u *User) createdInitialPolicyUser(tx *pop.Connection, householdID string) (bool, error) {
```
Schparky, line 339: Should this be `createInitialPolicyUser`? Otherwise, it sounds like it is a function with no side-effects.

application/models/user.go
```diff
@@ -324,6 +336,28 @@ func (u *User) CreateInitialPolicy(tx *pop.Connection, householdID string) error
...
+	var policy Policy
+	err := tx.Where("household_id = ?", householdID).First(&policy)
+	if domain.IsOtherThanNoRows(err) {
+		msg := fmt.Sprintf("error fetching policies with household id %s: %s", householdID, err.Error())
+		panic(msg)
+	}
+
+	if policy.ID != uuid.Nil {
```
Schparky, line 347:
```suggestion
	if policy.ID == uuid.Nil {
		return false, nil
	}	
```

Schparky (approved):

application/models/user.go
Baggerone: I basically didn't know what to call it, since it was potentially creating something and also returning a boolean to say whether it had.

Schparky: If it makes you feel better, you could return a `*PolicyUser` and leave it nil if it doesn't get created.

PR #347, opened 2021-12-10
set the income account for claims

Schparky:
Claims have a different "income" account number

### DIFF
--- a/application/domain/domain.go
+++ b/application/domain/domain.go
@@ -189,8 +189,8 @@ var Env struct {
 	EvacuationDeductible  float64 `default:"0.333333333" split_words:"true"`
 
 	FiscalStartMonth   int    `default:"1" split_words:"true"`
-	ExpenseAccount     string `default:"" split_words:"true"`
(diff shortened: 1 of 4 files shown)


Baggerone (approved):

application/domain/domain.go
```diff
@@ -188,8 +188,9 @@ var Env struct {
 	DeductibleString      string  `ignored:"true"`
 	EvacuationDeductible  float64 `default:"0.333333333" split_words:"true"`
 
-	FiscalStartMonth int    `default:"1" split_words:"true"`
-	ExpenseAccount   string `default:"" split_words:"true"`
+	FiscalStartMonth   int    `default:"1" split_words:"true"`
+	ExpenseAccount     string `default:"" split_words:"true"`
+	ClaimIncomeAccount string `default:"" split_words:"true"`
```
fillup, line 193: Should we also be setting required on vars that are required for proper operation? I think we'd rather the app exit on start with an error than run and have a problem generating charges because a var was missing.

Schparky: good idea

PR #350, opened 2021-12-10
remove TODOs related to steward/signator distinction

Schparky:
I believe that we have this resolved, simply by disallowing the same user from approving at Review2 and Review3. Is that sufficient?


mtompset (approved):
Removing comments. Assuming they are no longer relevant, this is an easy approval.

Baggerone (approved):

PR #355, opened 2021-12-14
Filter policies by username

Baggerone:


application/models/policy.go
```diff
@@ -282,12 +282,20 @@ func (p *Policies) Query(tx *pop.Connection, query api.Query) error {
...
-		return q.Where("policies.id IN (SELECT policies.id FROM policies,policy_users,users"+
-			" WHERE policies.id=policy_users.policy_id AND users.id = policy_users.user_id"+
-			" AND (users.first_name ILIKE ? OR users.last_name ILIKE ?"+
-			" OR policies.cost_center ILIKE ? OR policies.household_id ILIKE ?"+
-			" OR policies.name ILIKE ?))", searchText, searchText, searchText, searchText, searchText)
+		return q.Where("policies.id IN (SELECT policies.id "+
+			"FROM policies "+
+			"RIGHT JOIN policy_users pu ON policies.id = pu.policy_id "+
+			"RIGHT JOIN users on users.id = pu.user_id "+
+			"AND (CONCAT(users.first_name, ' ', users.last_name) ILIKE ? "+
```
jason-jackson, line 295: Shouldn't `AND` be `WHERE`?

application/models/policy.go
```diff
@@ -282,12 +282,20 @@ func (p *Policies) Query(tx *pop.Connection, query api.Query) error {
...
+			"OR policies.cost_center ILIKE ? OR policies.household_id ILIKE ? "+
+			"OR policies.name ILIKE ?) "+
+			"where policies.id is not null)", searchText, searchText, searchText, searchText)
```
jason-jackson, line 298: Capitalize `WHERE`

jason-jackson (approved):

application/models/policy.go
Baggerone: The `where` is outside the parenthesis, while the `and` is inside the parenthesis.

application/models/policy.go
Schparky: I had to reformat it to see what's going on. Maybe it would be good to make some edits to help the next person reading it. Here's what I ended up with, though you may not need to go this far (also note the capitalization on SQL keywords):
```sql
policies.id IN (
	SELECT policies.id FROM policies
		RIGHT JOIN policy_users pu ON policies.id = pu.policy_id
		RIGHT JOIN users ON users.id = pu.user_id
			AND (
				CONCAT(users.first_name, ' ', users.last_name) ILIKE ?
				OR policies.cost_center ILIKE ? OR policies.household_id ILIKE ?
				OR policies.name ILIKE ?
			)
		WHERE policies.id IS NOT NULL
)
```

Schparky: I think the `RIGHT` can be omitted since none of these foreign keys are nullable.

Schparky (approved):

PR #356, opened 2021-12-14
add calc totals to policy endpoints

RussellChamp:
still needs a little work. tell me why it sucks @Schparky 

Still needs 
* Ability to specify `GET /policies?totals` to include totals on that request
* Set the 'totals' object to null or something else when totals is not being populated

### DIFF
--- a/application/actions/policies.go
+++ b/application/actions/policies.go
@@ -33,10 +33,6 @@ import (
 //   in: query
 //   required: false
 //   description: comma-separated list of search pairs like "field:text". Presently, only meta-field 'active' is supported
-// - name: totals
-//   in: query
-//   required: false
-//   description: calculate and return policy totals
 // responses:
 //   '200':
 //     description: all policies
@@ -101,7 +97,7 @@ func policiesListCustomer(c buffalo.Context) error {
 func policiesView(c buffalo.Context) error {
 	policy := getReferencedPolicyFromCtx(c)
 
-	return renderOk(c, policy.ConvertToAPI(models.Tx(c), true, true))
+	return renderOk(c, policy.ConvertToAPI(models.Tx(c), true))
 }
 
 // swagger:operation POST /policies Policies PoliciesCreateTeam
@@ -144,7 +140,7 @@ func policiesCreateTeam(c buffalo.Context) error {
 		return reportError(c, err)
 	}
 
-	return renderOk(c, policy.ConvertToAPI(tx, true, true))
+	return renderOk(c, policy.ConvertToAPI(tx, true))
 }
 
 // swagger:operation PUT /policies/{id} Policies PoliciesUpdate
@@ -202,7 +198,7 @@ func policiesUpdate(c buffalo.Context) error {
 		return reportError(c, err)
 	}
 
-	return renderOk(c, policy.ConvertToAPI(tx, true, true))
+	return renderOk(c, policy.ConvertToAPI(tx, true))
 }
 
 // swagger:operation GET /policies/{id}/members PolicyMembers PolicyMembersList
--- a/application/actions/policies.go
+++ b/application/actions/policies.go
@@ -33,6 +33,10 @@ import (
 //   in: query
 //   required: false
 //   description: comma-separated list of search pairs like "field:text". Presently, only meta-field 'active' is supported
+// - name: totals
+//   in: query
+//   required: false
+//   description: calculate and return policy totals
 // responses:
 //   '200':
 //     description: all policies
@@ -97,7 +101,7 @@ func policiesListCustomer(c buffalo.Context) error {
 func policiesView(c buffalo.Context) error {
 	policy := getReferencedPolicyFromCtx(c)
 
-	return renderOk(c, policy.ConvertToAPI(models.Tx(c), true))
+	return renderOk(c, policy.ConvertToAPI(models.Tx(c), true, true))
 }
 
 // swagger:operation POST /policies Policies PoliciesCreateTeam
@@ -140,7 +144,7 @@ func policiesCreateTeam(c buffalo.Context) error {
 		return reportError(c, err)
 	}
 
-	return renderOk(c, policy.ConvertToAPI(tx, true))
+	return renderOk(c, policy.ConvertToAPI(tx, true, true))
 }
 
 // swagger:operation PUT /policies/{id} Policies PoliciesUpdate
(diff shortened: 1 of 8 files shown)


application/actions/policies.go
```diff
@@ -33,6 +33,10 @@ import (
 //   in: query
 //   required: false
 //   description: comma-separated list of search pairs like "field:text". Presently, only meta-field 'active' is supported
+// - name: totals
```
Schparky, line 36: I think this is not needed. We can just include the totals all of the time. I'm not sure performance is the driver for adding this. Our usual philosophy is that we don't solve performance problems until they are known to exist.

application/actions/policies.go
```diff
@@ -97,7 +101,7 @@ func policiesListCustomer(c buffalo.Context) error {
 func policiesView(c buffalo.Context) error {
 	policy := getReferencedPolicyFromCtx(c)
 
-	return renderOk(c, policy.ConvertToAPI(models.Tx(c), true))
+	return renderOk(c, policy.ConvertToAPI(models.Tx(c), true, true))
```
Schparky, line 104: While we have yet to do this, an interesting tactic I recently saw was to assign a variable to `bool` parameters just for passing into a function. Better for self-documenting code.

application/api/items.go
```diff
@@ -13,6 +13,10 @@ import (
 // swagger:model
 type ItemCoverageStatus string
 
+func (status ItemCoverageStatus) IsActive() bool {
+	return status == ItemCoverageStatusApproved
```
Schparky, line 17: If "Approved" isn't the right word, it's probably better to rename it than to wrap it in a function. If we want to fully wrap const values, we should try to do it universally rather than piecemeal. (Yes, I know I have added one or two of these. Confession is good for the soul.)

application/models/policy.go
```diff
@@ -221,7 +221,7 @@ func (p *Policy) LoadEntityCode(tx *pop.Connection, reload bool) {
 	}
 }
 
-func (p *Policy) ConvertToAPI(tx *pop.Connection, hydrate bool) api.Policy {
+func (p *Policy) ConvertToAPI(tx *pop.Connection, hydrate, calcTotals bool) api.Policy {
```
Schparky, line 224: As per my comment on the handler, I recommend removing this. Also, I'm slightly averse to adding another bool parameter. It makes the calling code a little ambiguous to read.

application/api/claims.go
```diff
@@ -49,6 +49,14 @@ func (s ClaimStatus) WasReviewed() bool {
 	return false
 }
 
+func (s ClaimStatus) IsOpen() bool {
+	switch s {
+	case ClaimStatusDraft, ClaimStatusReview1, ClaimStatusReview2, ClaimStatusReview3, ClaimStatusRevision, ClaimStatusReceipt:
```
jason-jackson, line 54: I feel like it might be easier to specify closed statuses here instead of open statuses.

application/api/claims.go
```diff
@@ -49,6 +49,14 @@ func (s ClaimStatus) WasReviewed() bool {
 	return false
 }
 
+func (s ClaimStatus) IsOpen() bool {
```
jason-jackson, line 52: I'd lean towards attaching this to claim and not claim status

Schparky (commented):
Good job on your first venture into this codebase!

application/api/items.go
```diff
@@ -13,6 +13,10 @@ import (
 // swagger:model
 type ItemCoverageStatus string
 
+func (status ItemCoverageStatus) IsActive() bool {
```
jason-jackson, line 16: I'd lean towards attaching this to item instead of itemcoveragestatus

application/models/claim.go
```diff
@@ -650,6 +650,19 @@ func (c *Claims) ByStatus(tx *pop.Connection, statuses []api.ClaimStatus) error
 	return appErrorFromDB(err, api.ErrorQueryFailure)
 }
 
+func (c *Claims) CalcTotals() api.ClaimTotals {
+	var claimTotals api.ClaimTotals
+
+	for _, claim := range *c {
+		if claim.Status.IsOpen() {
```
Baggerone, line 657: I wonder if `Draft` claims should really be included in this. Maybe `IsOpen()` should be `isPending()`

application/models/item.go
```diff
@@ -369,6 +369,19 @@ func (i *Item) IsActorAllowedTo(tx *pop.Connection, actor User, perm Permission,
 	return false
 }
 
+func (i *Items) CalcTotals() api.ItemTotals {
+	var itemTotals api.ItemTotals
+
+	for _, item := range *i {
+		if item.CoverageStatus.IsActive() {
```
Baggerone, line 376: Is there a drawback to just saying `if item.CoverageStatus == ItemCoverageStatusApproved`?

application/models/policy.go
Baggerone: If `calcTotals` is removed, would it be best to only calculate the totals if `hydrate` is true?

Baggerone (commented):
Before it's merged, it would be good to add to the tests for this.

application/actions/policies.go
RussellChamp: @Schparky imo a better way to do this is with meaningfully named enums (or consts if you want to support backward compatibility). that will force consumers into good behavior. in golang i believe that might look like
```
type HydrateChoice int
const (
  HydrateData HydrateChoice = 0
  NoHydrateData
)
```
then call
```
renderOk(c, p, HydrateData)
```

Schparky: That's a good suggestion. Another, more scalable method, is to use a variadic parameter that can take any number of options. It's more expensive to implement, but sure does make for a nice interface.

application/api/claims.go
RussellChamp: in that case, you have written a `IsNotClosed` method and not a `IsOpen` method

application/models/item.go
RussellChamp: in the case that we change the way an item's active-ness is changed, the code is all in one place. this also followed existing patterns.

application/api/items.go
RussellChamp: the partially defined desired work was to sum up active items. this may mean something different than implemented here and allows for all the code to be kept in one place if things need to be updated.

application/api/claims.go
RussellChamp: this follows current code patterns. should we update the other offending instances into conformity as well?

Schparky: Could actually do both. The `Claim` method would call the `ClaimStatus` method.

Schparky: What _should_ we do? I don't know. These details don't seem important at present.

Baggerone:
Outdated now

PR #363, opened 2021-12-21
rework ledger reporting

Schparky:
I learned that the ledger reporting doesn't happen on clean month-end boundaries, but should happen __prior__ to the end of the month so it can be entered into the accounting system __before__ the end of the month.


application/actions/ledger.go
```diff
@@ -36,49 +36,49 @@ func batchesGetLatest(c buffalo.Context) error {
...
-		err := fmt.Errorf("user not allowed to approve monthly batch data")
+		err := fmt.Errorf("user not allowed to reconcile ledger data")
 		return reportError(c, api.NewAppError(err, api.ErrorNotAuthorized, api.CategoryForbidden))
 	}
 
 	tx := models.Tx(c)
 
 	now := time.Now().UTC()
-	firstDay := domain.BeginningOfLastMonth(now)
+	date := domain.BeginningOfDay(now)
```
jason-jackson, line 79: My main concern is if this is done the day after the ledger is pulled, there may be some entries that are marked as reconciled, but were not actually checked.

Schparky: > My main concern is if this is done the day after the ledger is pulled, there may be some entries that are marked as reconciled, but were not actually checked.

Good observation. I added `end_date` to the request body, so the API client can provide the date of the last LedgerList request. Ideally, I'd like to make it more foolproof. I might do that before we make this accessible on the UI

application/actions/claims.go
```diff
@@ -341,7 +341,7 @@ func claimsRequestReceipt(c buffalo.Context) error {
 	return c.Render(http.StatusOK, r.JSON(output))
 }
 
-// swagger:operation POST /claims/{id}/approve Claims ClaimsApprove
+// swagger:operation POST /claims/{id} Claims ClaimsApprove
```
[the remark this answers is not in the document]
Schparky: I think that was accidental. I didn't intend to make any changes to `/claims`

fillup (approved):

PR #365, opened 2021-12-22
ledger corrections

Schparky:
fixed a few more ledger reporting issues:
- fiscal year wasn't correct
- CSV reference column was missing the entity code and account number
- (minor) description column had redundant information

### DIFF
--- a/application/fin/fin_test.go
+++ b/application/fin/fin_test.go
@@ -2,6 +2,7 @@ package fin
 
 import (
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/assert"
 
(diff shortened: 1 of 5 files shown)


application/fin/fin_test.go
```diff
@@ -33,3 +34,38 @@ func Test_getFiscalPeriod(t *testing.T) {
...
+			date:        time.Date(2021, 8, 31, 0, 0, 0, 0, time.UTC),
+			want:        2021,
+		},
+		{
+			name:        "fiscal start in January",
+			fiscalStart: 1,
+			date:        time.Date(2021, 1, 1, 0, 0, 0, 0, time.UTC),
+			want:        2021,
+		},
+	}
```
fillup, line 63: The math looks good above, but since it'd be easy and cheap how about having a test for some day in every month of the year to ensure it's getting the right fiscal year value?

fillup (approved):

PR #367, opened 2021-12-22
Release - ledger reporting corrections

Schparky:


fillup (approved):

application/models/ledgerentry.go
```diff
@@ -169,28 +170,27 @@ func (le *LedgerEntry) Reconcile(ctx context.Context, now time.Time) error {
 	return nil
 }
 
-// TODO: make a better description format unless it has to be the same as before (which I doubt)
 func (le *LedgerEntry) transactionDescription() string {
-	dateString := le.DateSubmitted.Format("Jan 02, 2006")
+	dateString := le.DateSubmitted.Format("01-02-2006")
```
jason-jackson, line 174: This might be better as a constant / env variable. (I don't think it needs to be changed right away)

jason-jackson (approved):

PR #373, opened 2022-01-04
policies pagination

RussellChamp:
Display pagination options from the policies list endpoint

### DIFF
--- a/application/api/query_test.go
+++ b/application/api/query_test.go
@@ -12,6 +12,7 @@ func (ts *TestSuite) TestNewQuery() {
 		name             string
 		qs               string
 		wantLimit        int
+		wantPage         int
 		wantFilterActive string
 		wantSearchText   string
 	}{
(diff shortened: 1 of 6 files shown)


application/models/policy.go
```diff
@@ -262,22 +262,24 @@ func (p *Policies) All(tx *pop.Connection) error {
...
 }
 
-func (p *Policies) Query(tx *pop.Connection, query api.Query) error {
+func (p *Policies) Query(tx *pop.Connection, params api.QueryParams) (*pop.Query, error) {
 	q := tx.Order("updated_at DESC")
 
-	if query.Limit() > 0 {
-		q.Limit(query.Limit())
+	if params.Limit() > 0 && params.Page() > 0 {
+		q.Paginate(params.Page(), params.Limit())
```
Schparky, line 269: Can these two conditions be combined? For example:
```
if params.Limit() > 0 {
	q.Paginate(params.Page(), params.Limit())
}
```
Just need to ensure page is never < 1

application/models/policy.go
```diff
@@ -262,22 +262,24 @@ func (p *Policies) All(tx *pop.Connection) error {
 	return appErrorFromDB(tx.All(p), api.ErrorQueryFailure)
 }
 
-func (p *Policies) Query(tx *pop.Connection, query api.Query) error {
+func (p *Policies) Query(tx *pop.Connection, params api.QueryParams) (*pop.Query, error) {
```
Schparky, line 265: Why did you decide to return the `Query` when all that's needed is the `Paginator`? I might even suggest defining a custom type in the `api` package and transfer the useful data from `pop.Paginator` into that, to minimize coupling to the buffalo package.

Schparky (changes requested):

application/models/policy.go
RussellChamp: i didn't want to change existing functionality but yes that would simplify things

application/models/policy.go
RussellChamp: it seemed to be a more reusable/future-extensible pattern and made more sense than returning a Paginator from a `Query` method. we're already taking in a `*pop.Connection` as a parameter so we seem pretty bound to that implementation already. is this a soft or hard suggestion?

application/api/query_test.go
```diff
@@ -44,7 +44,7 @@ func (ts *TestSuite) TestNewQuery() {
 		ts.T().Run(tt.name, func(t *testing.T) {
 			values, _ := url.ParseQuery(tt.qs)
 
-			got := NewQuery(buffalo.ParamValues(values))
+			got := NewQueryParams(buffalo.ParamValues(values))
```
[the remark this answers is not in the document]
RussellChamp: yes

application/models/policy.go
Schparky: Good point about the `pop.Connection` parameter. The confusing thing to me, though, is from a consumer perspective (or just a person reading the code) one might think that the actual query hasn't taken place since you're returning the `pop.Query`. So the question is, do I need to call `q.Exec` on the returned `Query`? The suggestion to create a new type is "soft" but I'm still unsure about the use of `Query`. Want to talk directly?

Schparky (approved):

PR #374, opened 2022-01-06
Release

Schparky:
### Added
- Add pagination on steward/signator `PolicyList` endpoint
- Create policy history and claim history records when policies, items, and claims are created
### Changed
- Sort claims, items, and dependents when associated with a policy


RussellChamp (approved):
lgtm

PR #377, opened 2022-01-14
Block policy invite creation/acceptance when different household id's are involved

Baggerone:
https://trello.com/c/C2Dj8cSi/466-household-policy-invitation

### DIFF
--- a/application/models/policyuserinvite.go
+++ b/application/models/policyuserinvite.go
@@ -129,6 +129,13 @@ func (i *PolicyUserInvite) Accept(tx *pop.Connection, code string, user User) er
 
 	i.LoadPolicy(tx, true)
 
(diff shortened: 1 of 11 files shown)


application/models/policyuserinvite.go
```diff
@@ -129,6 +129,13 @@ func (i *PolicyUserInvite) Accept(tx *pop.Connection, code string, user User) er
 
 	i.LoadPolicy(tx, true)
 
+	if i.Policy.HouseholdID.Valid && user.StaffID.Valid {
+		householdID := GetHHID(user.StaffID.String)
```
Schparky, line 133: The response time on the API call is very slow (several seconds). But I can't think of another way to do it at the moment, so could you add a Trello card to discuss possibilities?

Schparky (approved):
that was a lot more complicated than I had thought. Good work

PR #380, opened 2022-01-18
Add link and address to email footer

Baggerone:


application/templates/mail/_customer_footer.plush.html
```diff
@@ -23,9 +23,15 @@
...
 	<div style="background-color: #ebeef2; padding: 16px;">
 		<div style="font-size: 9px; margin-bottom: 8px;">
-			<%= appName %> is a service of SIL International.
+			<span style="margin-right: 10px;">
+				<a href="<%= uiURL %>/home">Your Account</a>
+			</span>
 		</div>
 		<div style="font-size: 9px; margin-bottom: 8px;">
+			<%= appName %> is a service of SIL International.<br>
+			SIL International, 7500 W Camp Wisdom Rd, Dallas, TX 75236, USA
```
Schparky, line 32: I think it's strange that the mockups have this address, when the sil.org site doesn't even give this information. If we really want it here, maybe put it in an environment variable.

Baggerone: The sil.org site does give the information on the `contact` page.

If we don't limit the org name to an environment variable, I doubt it's worth adding one for the address.

Schparky: Oh, sorry. Not sure how I missed that.

Schparky (approved):

application/templates/mail/_customer_footer.plush.html
```diff
@@ -23,9 +23,15 @@
 	</div>
 	<div style="background-color: #ebeef2; padding: 16px;">
 		<div style="font-size: 9px; margin-bottom: 8px;">
```
jason-jackson, line 25: If all of these internal divs have `font-size: 9px;`, why not put it in parent div style? (or is it an email thing?)

jason-jackson (approved):

PR #388, opened 2022-01-26
change S3 ACL to private

Schparky:
Previously, the S3 ACL was only being set to `private` for dev and test. It is now configurable, with `private` as the default. If set to a value other than `public-read` or `public-read-write`, pre-signed URLs with a configurable lifetime will be provided to the API consumer. The default lifetime is 10 minutes.

### DIFF
--- a/application/domain/domain.go
+++ b/application/domain/domain.go
@@ -165,14 +165,18 @@ var Env struct {
 	SamlRequireEncryptedAssertion   bool   `default:"true" split_words:"true"`
 
 	AwsRegion          string `split_words:"true"`
-	AwsS3Endpoint      string `split_words:"true"`
-	AwsS3DisableSSL    bool   `split_words:"true"`
-	AwsS3Bucket        string `split_words:"true"`
 	AwsAccessKeyID     string `split_words:"true"`
 	AwsSecretAccessKey string `split_words:"true"`
-	EmailFromAddress   string `required:"true" split_words:"true"`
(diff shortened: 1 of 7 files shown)


application/domain/domain.go
```diff
@@ -165,14 +165,18 @@ var Env struct {
...
 	AwsSecretAccessKey string `split_words:"true"`
-	EmailFromAddress   string `required:"true" split_words:"true"`
-	EmailService       string `default:"ses" split_words:"true"`
-	SupportEmail       string `default:"" split_words:"true"`
+
+	AwsS3Endpoint       string `split_words:"true"`
+	AwsS3DisableSSL     bool   `split_words:"true"`
+	AwsS3Bucket         string `split_words:"true"`
+	AwsS3ACL            string `default:"private" split_words:"true"`
+	AwsS3URLLifeMinutes int    `default:"10" split_words:"true"`
```
jason-jackson, line 175: Could we use `AwsS3UrlLifeMinutes` instead?

jason-jackson (approved):

application/domain/domain.go
Schparky: Though we're not 100% consistent, we have been using `URL` more than `Url`

Schparky: I generally like to capitalize short acronyms even in camel and Pascal case identifiers, but in this case, I think I agree that it doesn't read well.

PR #393, opened 2022-01-28
group renewals by policy (not one transaction per item)

Schparky:


application/actions/ledger.go
```diff
@@ -163,7 +163,11 @@ func ledgerAnnualProcess(c buffalo.Context) error {
 
 	currentYear := time.Now().UTC().Year()
 
-	if err := models.ProcessAnnualCoverage(tx, currentYear); err != nil {
+	var policies models.Policies
+	if err := policies.AllActive(tx); err != nil {
+		return err
```
Baggerone, line 168: Should this be `return reportError(c, err)`?
(If so, it looks like there might be a couple of other places in the file that may need to change.)

application/models/policy.go
```diff
@@ -528,3 +532,64 @@ func (p *Policy) createInvite(tx *pop.Connection, invite api.PolicyUserInviteCre
 	}
 	return nil
 }
+
+// ProcessAnnualCoverage creates coverage renewal ledger entries for all items covered for the given year.
+// Does not create new records for items already processed.
+func (p *Policies) ProcessAnnualCoverage(tx *pop.Connection, year int) error {
+	for _, pp := range *p {
+		if err := pp.ProcessAnnualCoverage(tx, year); err != nil {
+			return fmt.Errorf("error processing annual coverage for policy %s: %w", pp.ID, err)
```
Baggerone, line 541: I'm wondering about the mix of returning normal errors and AppErrors (lines 553 and 560).
Are you OK with that?

Baggerone (approved):
A couple of things to consider.

application/actions/ledger.go
Schparky: Thanks for spotting these

application/models/policy.go
Schparky: Thanks. Might have been ok, but much better to be consistent.

PR #396, opened 2022-01-31
implement LedgerReport

Schparky:
This provides the assurance that when reconciling, only the transactions included in the report will be reconciled. Previously, the client was trusted to pass in the same date that was used to create the report originally. 

Note: the reconcile endpoint isn't finished yet, so it still works as if the LedgerReport type doesn't exist.

### DIFF
--- a/application/models/ledgerreport.go
+++ b/application/models/ledgerreport.go
@@ -121,6 +121,11 @@ func NewLedgerReport(ctx context.Context, reportType string, date time.Time) (Le
 
 	report := LedgerReport{Type: reportType}
 
+	if date.After(time.Now().UTC()) {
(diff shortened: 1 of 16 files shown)


application/models/ledgerreport.go
```diff
@@ -0,0 +1,159 @@
...
+			return report, err
+		}
+	case ReportTypeAnnual:
+		year := date.Year()
+		report.Date = time.Date(year, 1, 1, 0, 0, 0, 0, time.UTC)
+		if err := le.FindRenewals(tx, year); err != nil {
+			return report, err
+		}
+	default:
+		err := errors.New("invalid report type")
```
Baggerone, line 143: It might be worth adding the value of the type here to make it clear.

Baggerone (approved):

PR #402, opened 2022-02-01
move files to a new path: {id}/{filename}

Schparky:
This helps the UI/UX because the browser offers the default filename, when downloading, as the filename in the URL. There's a cumbersome workaround available on the UI, but this  will be a cleaner solution.

### DIFF
--- a/application/main.go
+++ b/application/main.go
@@ -40,6 +40,12 @@ func main() {
 		os.Exit(1)
 	}
 
+	// temporary job to migrate S3 files to new path
+	if err := job.SubmitDelayed(job.MigrateFiles, delay+time.Second, map[string]interface{}{}); err != nil {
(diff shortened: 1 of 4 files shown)


application/main.go
```diff
@@ -40,6 +40,12 @@ func main() {
 		os.Exit(1)
 	}
 
+	// temporary job to migrate S3 files to new path
+	if err := job.SubmitDelayed(job.MigrateFiles, delay+time.Second, map[string]interface{}{}); err != nil {
```
Baggerone, line 44: I'm wondering if you want a different (shorter) delay for this.

Baggerone (approved):

application/main.go
Schparky: Good call.

PR #407, opened 2022-02-02
ledger report reconcile

Schparky:


application/models/ledgerreport.go
```diff
@@ -91,14 +91,25 @@ func (lr *LedgerReport) IsActorAllowedTo(tx *pop.Connection, actor User, perm Pe
 // ConvertToAPI converts a LedgerReport to api.LedgerReport
 func (lr *LedgerReport) ConvertToAPI(tx *pop.Connection) api.LedgerReport {
```
Baggerone, line 92: (This is an older issue, but maybe it's worth changing now.)

The matching method for LedgerReports (on line 34) returns an interface.  Is that necessary? I think best-practice in Go is to accept interfaces but return concrete types.

application/models/ledgerreport.go
```diff
@@ -91,14 +91,25 @@ func (lr *LedgerReport) IsActorAllowedTo(tx *pop.Connection, actor User, perm Pe
...
 	lr.LoadFile(tx, false)
+	lr.LoadLedgerEntries(tx, false)
+
+	transactionCount := len(lr.LedgerEntries)
+	unclearedCount := 0
+	for _, e := range lr.LedgerEntries {
+		if !e.DateEntered.Valid {
+			unclearedCount++
+		}
```
Baggerone, line 101: Would it be OK to add a `break` statement here? 
Or were you planning to use the full `unclearedCount` somehow?

Baggerone (approved):
A couple of suggestions.

application/models/ledgerreport.go
Schparky: I think that may be an artifact of using Go's "implement function" feature. I sometimes forget to give it a specific return type. Thanks for drawing it to my attention.

application/models/ledgerreport.go
Schparky: Good one. I'll even change the variable to a bool instead of an int.

PR #412, opened 2022-02-24
create new policy ledger report

Baggerone:
Am I on the right track here? (More to do later.)

### DIFF
--- a/application/actions/policies_test.go
+++ b/application/actions/policies_test.go
@@ -648,7 +648,6 @@ func (as *ActionSuite) Test_PolicyLedgerReportCreate() {
 				Month: tt.month,
 				Year:  tt.year,
 				Type:  tt.reportType,
-				//Date: time.Now().UTC().Format(domain.DateFormat),
 			})
 
 			body := res.Body.String()
--- a/application/actions/policies_test.go
+++ b/application/actions/policies_test.go
@@ -89,7 +89,7 @@ func (as *ActionSuite) Test_PoliciesList() {
 		as.T().Run(tt.name, func(t *testing.T) {
 			req := as.JSON("/policies" + tt.queryString)
 			req.Headers["Authorization"] = fmt.Sprintf("Bearer %s", tt.actor.Email)
-			req.Headers["content-type"] = "application/json"
+			req.Headers["content-type"] = domain.ContentJson
 			res := req.Get()
 
 			body := res.Body.String()
@@ -178,7 +178,7 @@ func (as *ActionSuite) Test_PoliciesView() {
 		as.T().Run(tt.name, func(t *testing.T) {
 			req := as.JSON("/policies/" + tt.policyID.String())
 			req.Headers["Authorization"] = fmt.Sprintf("Bearer %s", tt.actor.Email)
-			req.Headers["content-type"] = "application/json"
+			req.Headers["content-type"] = domain.ContentJson
 			res := req.Get()
 
 			body := res.Body.String()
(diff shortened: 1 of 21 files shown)


application/models/ledgerentry.go
```diff
@@ -94,6 +96,30 @@ func (le *LedgerEntries) AllNotEntered(tx *pop.Connection, cutoff time.Time) err
...
+
+	var buf bytes.Buffer
+	buf.Write([]byte(csvPolicyHeader))
+
+	for _, l := range *le {
+		if l.Amount == 0 {
+			continue
+		}
+
+		nextRow := fmt.Sprintf(
```
Schparky, line 110: I don't remember why I didn't use the `encoding/csv` package. If you're interested, it might be worth another look.

application/domain/domain.go
```diff
@@ -89,16 +89,19 @@ const (
...
 	// How many hours old an item can be until it's not allowed to be deleted
 	ItemDeleteCutOffHours = 72
 
 	MaxFileSize = 1024 * 1024 * 10 // 10 Megabytes
+	Megabyte    = 1048576
 
-	DurationDay  = time.Duration(time.Hour * 24)
-	DurationWeek = time.Duration(DurationDay * 7)
-	Megabyte     = 1048576
+	TextCSV = "text/csv"
```
Schparky, line 104: There are several other content types later in this file. If you want to use `const` for them, feel free.

application/models/ledgerreport.go
```diff
@@ -173,12 +173,95 @@ func NewLedgerReport(ctx context.Context, reportType string, date time.Time) (Le
...
+//   to the requested report type and the month and year of the request
+func NewPolicyLedgerReport(ctx context.Context, policy Policy, reportType string, month, year int) (LedgerReport, error) {
+	tx := Tx(ctx)
+
+	now := time.Now().UTC()
+	nowYear := now.Year()
+
+	report := LedgerReport{Type: reportType}
+
+	if year < 2000 || year > nowYear {
```
Schparky, line 192: 2000 seems arbitrary. I think there was some activity prior to that.

Schparky (commented):
Looks good to me.

application/models/ledgerreport.go
```diff
@@ -173,12 +173,95 @@ func NewLedgerReport(ctx context.Context, reportType string, date time.Time) (Le
...
+	report := LedgerReport{Type: reportType}
+
+	if year < 1971 || year > nowYear {
+		err := fmt.Errorf("invalid year requested: %d", year)
+		return report, api.NewAppError(err, api.ErrorInvalidDate, api.CategoryUser)
+	}
+
+	var le LedgerEntries
+	switch reportType {
+	case ReportTypeMonthly:
```
Schparky, line 199: Thinking it might be better to use different enum values for one-policy reports. Having them mixed up in the database might be confusing. Along these lines, maybe add `policy_id` to the `ledger_reports` table to identify which policy it's for. Technically, you could grab the policy ID from any of the `ledger_item` records on the report, so maybe not totally necessary.

application/actions/policies_test.go
```diff
@@ -577,11 +578,93 @@ func (as *ActionSuite) Test_PoliciesInviteMember() {
...
+
+	for _, tt := range tests {
+		as.T().Run(tt.name, func(t *testing.T) {
+			req := as.JSON("%s/%s/ledger-reports", policiesPath, policy.ID.String())
+			req.Headers["Authorization"] = fmt.Sprintf("Bearer %s", tt.actor.Email)
+			res := req.Post(api.PolicyLedgerReportCreateInput{
+				Month: tt.month,
+				Year:  tt.year,
+				Type:  tt.reportType,
+				//Date: time.Now().UTC().Format(domain.DateFormat),
```
Schparky, line 651: Delete?

application/models/ledgerreport.go
Baggerone: How about if I just add a nullable `policy_id` column to the `ledger_reports` table?  Then I wouldn't need to use different enum values for the `reportType`, right?

Schparky: Yes, that would work

PR #415, opened 2022-02-28
Protect ledger reports that have a policy

Baggerone:


application/models/ledgerreport.go
```diff
@@ -138,6 +145,15 @@ func (lr *LedgerReport) LoadLedgerEntries(tx *pop.Connection, reload bool) {
 	}
 }
 
+// LoadPolicy hydrates the Policy property if necessary or if `reload` is true.
+func (lr *LedgerReport) LoadPolicy(tx *pop.Connection, reload bool) {
+	if lr.Policy.ID == uuid.Nil || reload {
```
Schparky, line 150: Should this check `policy_id` for `NULL`? I suppose it's redundant with the one instance it's used, but it would future-proof it.
```suggestion
	if (lr.PolicyID.Valid && lr.Policy.ID == uuid.Nil) || reload {
```

Schparky (approved):

PR #416, opened 2022-03-01
csv encoding for ledger reports

Baggerone:
I'm not sure if this is an improvement.
Does it matter that multiple 0's at the beginning of a value don't get double quotes around them? 

Steve, if you think this looks OK, can you test the admin ledger entries manually?

### DIFF
--- a/application/fin/sage.go
+++ b/application/fin/sage.go
@@ -2,21 +2,18 @@ package fin
 
 import (
 	"bytes"
+	"encoding/csv"
+	"errors"
 	"fmt"
 
 	"github.com/silinternational/cover-api/api"
 )
 
-const (
-	header1 = `"RECTYPE","BATCHID","BTCHENTRY","ORIGCOMP","SRCELEDGER","SRCETYPE","FSCSYR","FSCSPERD","SWEDIT",` +
-		`"JRNLDESC","REVPERD","ERRBATCH","ERRENTRY","DETAILCNT","PROCESSCMD"` + "\n"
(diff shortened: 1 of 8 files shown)


application/fin/sage.go
```diff
@@ -32,33 +29,73 @@ func (s *Sage) AppendToBatch(t Transaction) {
...
+
 	for i := range s.Transactions {
-		buf.Write(s.transactionRow(i))
+		if err := writer.Write(s.transactionRow(i)); err != nil {
+			return []byte{}, errors.New("error writing sage batch to csv: " + err.Error())
+		}
 	}
 
-	return buf.Bytes()
+	writer.Flush()
```
jason-jackson, line 50: Not sure this is needed as there's a defer already.

application/models/ledgerentry.go
```diff
@@ -96,33 +99,44 @@ func (le *LedgerEntries) AllNotEntered(tx *pop.Connection, cutoff time.Time) err
...
-		)
-		buf.Write([]byte(nextRow))
+		}
+
+		if err := writer.Write(nextRow); err != nil {
+			return []byte{}, errors.New("error adding to Ledger Entry csv output: " + err.Error())
+		}
+	}
+
+	writer.Flush()
```
jason-jackson, line 128: Not sure this is needed as there's a defer Flush already.

Baggerone:
The csv encoding is too bulky to be worth it, plus it doesn't deal with quoting zeros very well.

PR #418, opened 2022-03-08
Return policies with their user invites

Baggerone:


application/api/policyuserinvites.go
```diff
@@ -0,0 +1,20 @@
...
+	// invitee's email
+	Email string `json:"email"`
+
+	// invitee's name
+	Name string `json:"name"`
+
+	// date and time when invite email was sent (omitted if empty)
+	//
+	// swagger:strfmt date-time
+	EmailSentAt *time.Time `json:"email_sent_at,omitempty"`
```
Schparky, line 19: Do we keep record of whether the person responded to the invite? Seems like that might be useful on the UI.

Schparky (approved):

application/api/policyuserinvites.go
Baggerone: The invite gets destroyed once the new user accepts it.

PR #419, opened 2022-03-08
delete a claim

Baggerone:


Schparky (commented):
Can you add a boolean property to claims that indicates whether it can be deleted? That way the UI code doesn't have to duplicate the business logic.

Schparky (approved):

PR #420, opened 2022-03-09
Make one function responsible for deciding whether a claim is deletable

Baggerone:


jason-jackson (approved):

mtompset (approved):
I'm assuming @Baggerone ran this. It looks like a simple function extract and tiny refactor.

Schparky (approved):

PR #431, opened 2022-03-18
don't create ledger entries the value of which would be zero

Baggerone:


Schparky (approved):
I had considered leaving these in place just as a record of the transaction. I believe report generation filters these out.

Baggerone:
@Schparky 
I was just obeying the trello card. Should I remove the `api` label from it?
https://trello.com/c/69WbDzfo/469-ignore-0-transactions

PR #432, opened 2022-03-18
minimum charge of $1

Baggerone:


application/models/item.go
```diff
@@ -924,11 +924,12 @@ func (i *Item) SetAccountablePerson(tx *pop.Connection, id uuid.UUID) error {
 	return api.NewAppError(errors.New("accountable person ID not found"), api.ErrorNoRows, api.CategoryUser)
 }
 
-// CreateLedgerEntry does nothing if the amount is zero
+// CreateLedgerEntry creates a charge of at least $1
 func (i *Item) CreateLedgerEntry(tx *pop.Connection, entryType LedgerEntryType, amount api.Currency) error {
-	if amount == 0 {
-		return nil
+	if amount < 100 { // Charge at least $1
```
Schparky, line 929: Need to check `le.Type` for `LedgerEntryTypeCoverageRefund` and use `amount > -100` instead (or maybe `amount == 0` in this case?)

Schparky (changes requested):

Schparky (approved):

PR #434, opened 2022-03-30
Include a policy's ledger reports on its view

Baggerone:


application/models/ledgerreport.go
```diff
@@ -28,11 +28,25 @@ var ValidLedgerReportTypes = map[string]struct{}{
 
 type LedgerReports []LedgerReport
 
+func (lr LedgerReports) Create(tx *pop.Connection) error {
```
Schparky, line 31: This appears to the first of its kind (`Create` on a list type) and only used for tests. Should `testutils` do the iteration as it does for other types?

Schparky (commented):
Looks like a good change. Added some design decision comments for you to review.

application/models/policy.go
```diff
@@ -277,12 +277,7 @@ func (p *Policy) ConvertToAPI(tx *pop.Connection, hydrate bool) api.Policy {
 	}
 
 	if hydrate {
-		p.LoadClaims(tx, true)
-		p.LoadDependents(tx, true)
-		p.LoadInvites(tx, true)
-		apiPolicy.Claims = p.Claims.ConvertToAPI(tx)
-		apiPolicy.Dependents = p.Dependents.ConvertToAPI()
-		apiPolicy.Invites = p.Invites.ConvertToAPI()
+		p.hydrateAPiPolicy(tx, &apiPolicy)
```
Schparky, line 280: At first I thought the lowercase `i` was a typo, but then realize it's to separate words. Another name choice would be `hydratePolicyForAPI` with the acronym at the end to avoid the awkward casing. I'm fine with this, though, if you want to leave it as is.

application/models/policy.go
```diff
@@ -296,20 +291,23 @@ func (p *Policy) ConvertToAPI(tx *pop.Connection, hydrate bool) api.Policy {
...
+	p.LoadDependents(tx, true)
+	p.LoadInvites(tx, true)
+	apiPolicy.Claims = p.Claims.ConvertToAPI(tx)
+	apiPolicy.Dependents = p.Dependents.ConvertToAPI()
+	apiPolicy.Invites = p.Invites.ConvertToAPI()
+
 	var reports LedgerReports
 	if err := reports.AllForPolicy(tx, p.ID); domain.IsOtherThanNoRows(err) {
-		msg := fmt.Sprintf("error retrieving ledger reports for policy %s: %s", p.ID.String(), err)
+		msg := fmt.Sprintf("error retrieving recent strikes for policy %s: %s", p.ID.String(), err)
```
Schparky, line 304: It seems that strikes aren't loaded, unless I'm missing something in the diff.

Schparky: Oh, I see. It's still in `ConvertToAPI`. Should it be in the hydrate function?

Schparky, line 304:
```suggestion
		msg := fmt.Sprintf("error retrieving ledger reports for policy %s: %s", p.ID.String(), err)
```

Schparky (approved):

PR #437, opened 2022-04-13
Add faq url for email footer

Baggerone:


jason-jackson (approved):

codeship-services.yml
```diff
@@ -27,6 +27,7 @@ app:
     EMAIL_FROM_ADDRESS: no_reply@example.com
     EMAIL_SERVICE: dummy
     SUPPORT_URL: https://help.example.com
+    FAQ_URL: https://help.example.com/faq
```
Schparky, line 30: To avoid yet another required variable, can you make the default value equal to the support URL + "/faq"? Every time we add a new required environment variable, it has to be added in several places: codeship.yml, test.env, and GoLand run configuration are the three that I see at the moment. Optional variables could even be omitted from Terraform if the default is correct for our implementation.

Schparky (approved):
Approved, but with a minor request.

codeship-services.yml
Baggerone: The way it's set up, it's not a required variable, except for the tests. Do you want me to change the tests to give it a temporary value?

(Adding `/faq` doesn't match our faq url.)

Schparky: > The way it's set up, it's not a required variable, except for the tests. Do you want me to change the tests to give it a temporary value?
> 
> (Adding `/faq` doesn't match our faq url.)

Oh, sorry, I missed that. I think what you've done is fine, though it wouldn't be a problem to set the variable as part of the test setup. Another simplification might be to use an `env_file` directive to include `test.env` in the `codeship-services.yml` file.

PR #439, opened 2022-04-18
Create a draft replacement item when a claim item has been replaced

Baggerone:


application/models/claim.go
```diff
@@ -1061,3 +1061,53 @@ func (c *Claim) StopItemCoverage(tx *pop.Connection) error {
 
 	return nil
 }
+
```
Schparky, line 1064: Looking at the line number (1064) I'm reminded of a recent Go Time podcast. When someone mentioned that they _started_ to think about refactoring when their source file reached 1000 lines, the reactions from the other podcast participants were quite telling. He may as well have admitted to being a child molester. Well, that's an exaggeration, but point being that the consensus was that 1,000 lines is too long for a single file. What to do about it? I have no idea. It just adds to my building anxiety around management of legacy code.

application/models/claim.go
```diff
@@ -1061,3 +1061,53 @@ func (c *Claim) StopItemCoverage(tx *pop.Connection) error {
...
+		startDate := time.Now().UTC()
+		if ci.Item.CoverageEndDate.Valid {
+			startDate = ci.Item.CoverageEndDate.Time
+		}
+
+		newItem := Item{
+			Name:              ci.Item.Name,
+			CategoryID:        ci.Item.CategoryID,
+			RiskCategoryID:    ci.Item.RiskCategoryID,
+			InStorage:         false,
```
Schparky, line 1093: I didn't remember we had an `InStorage` field. Is it used? Why is it explicitly set to `false`?

Schparky (approved):
Looks like a good implementation. Approved, given that it's a necessary feature. (See Slack comment/question.)

application/models/claim.go
mtompset: If there is a way to extract some code to something else externally in a logical, maintainable way, then do it. Learn from the horror that is Mail Admin. :)

application/listeners/claim.go
```diff
@@ -114,17 +114,31 @@ func claimApproved(e events.Event) {
 		return
 	}
 
+	coverageStopped := false
+
 	models.DB.Transaction(func(tx *pop.Connection) error {
```
jason-jackson, line 119: What should happen if something is repaired? Keep coverage with different estimate?

application/models/claim.go
Baggerone: I'm having trouble understanding the cost-benefit ratio.  If we have shorter files, we get more files that we would have to navigate.  With our IDE's allowing us to jump here and there within the code and see where in a file a change has been made (etc), it seems like the file size isn't all that important.

application/models/claim.go
Baggerone: I don't know if it's really used yet. I think eventually they thought they would use it in some kind of values/calculations/limitations.

application/listeners/claim.go
Baggerone: If it's just repaired, we leave it to the user to modify the current item for now.

application/listeners/claim.go
```diff
@@ -114,17 +114,31 @@ func claimApproved(e events.Event) {
...
 		if err := claim.StopItemCoverage(tx); err != nil {
```
jason-jackson, line 120: Should this return a bool for coverage stopped?

application/models/claim.go
Schparky: The size/number of files is only a symptom. I think the root issue is the structure of the code. I'm thinking of architectural principles like DRY, KISS, SOLID, etc.

application/listeners/claim.go
Baggerone: We don't control that function or the type of the return value of the function param.

Schparky: @jason-jackson were you referring to `StopItemCoverage`?

Baggerone: What would be the benefit of having `StopItemCoverage` return a boolean? I don't think that value would get back to the calling function.

jason-jackson: Yeah, I don't think it would either, if we wanted it to, we'd have to rework it some more.

Schparky: Looks like it's connected to the conversation on Slack. I think your idea to rename `coverageStopped` to `coverageError` would be helpful if not adding a bool return value to `StopItemCoverage`. To summarize my understanding, to see if I'm tracking: `StopItemCoverage` returns `nil` if coverage was stopped or if no items were payout type "replace". The `if coverageStopped` statement a few lines down from here is not truly testing whether coverage was stopped since `coverageStopped` is set to `true` any time `StopItemCoverage` returns nil. It is not currently a problem but could be if the content of the `if` block changed.

Baggerone: I guess the way I see it is that `StopItemCoverage` returns nil if there were no errors. If there are errors, then the db transaction gets reversed, I'm guessing.

I suppose we could have `StopItemCoverage` return (bool, error) and then move the call to `DraftReplacmentItems` up into the same db transaction as the call to `StopItemCoverage`.

I was thinking it would be better to separate them out and let the email notification get created first, but maybe that doesn't matter.

jason-jackson: Yeah, that's roughly what I'm seeing, I'd be in favor of renaming the variable if this isn't renamed as, like you said, it's not really checking if coverage is stopped, but if there was an error in changing coverage.

jason-jackson: If the items were created first, they could possibly be used in the notification and that might be the better way to do it.

Schparky (approved):

PR #442, opened 2022-04-27
Remove team member from policy

Baggerone:


application/api/policymembers.go
```diff
@@ -35,4 +35,9 @@ type PolicyMember struct {
 
 	// a country
 	Country string `json:"country,omitempty"`
+
+	// unique ID
```
Schparky, line 39:
```suggestion
	// user ID of policy member
```

application/models/policyuser.go
```diff
@@ -62,3 +67,53 @@ func (p *PolicyUser) IsActorAllowedTo(tx *pop.Connection, actor User, perm Permi
 
 	return false
 }
+
+func (p *PolicyUser) FindByPolicyAndUserIDs(tx *pop.Connection, policyID, userID uuid.UUID) error {
+	return tx.Where(`policy_id = ? AND user_id = ?`, policyID, userID).First(p)
+}
+
+// Delete ensures there is an additional PolicyUser for the related policy and
```
Schparky, line 75:
```suggestion
// Delete removes a policy member if there is an additional PolicyUser for the related policy and
```

Schparky (commented):
If possible, I think it would be better to use the UserID instead of the PolicyUserID

Schparky:
> If possible, I think it would be better to use the UserID instead of the PolicyUserID

As you pointed out in Slack, PolicyUserID is consistent with my earlier suggestions.

Schparky (approved):

PR #444, opened 2022-04-29
Make alert colors match Figma examples

Baggerone:
Also reverted previous color change of the umbrella alert icon


Baggerone:
Example (not the red and pink colors)

image

application/templates/mail/_alert.plush.html
```diff
@@ -1,35 +1,67 @@
 
-<div style="margin:16px 0px; padding:8px; border-radius: 8px; background: #EBEEF2;">
+<%
+  let bgColor = "#EEEEEE"
+  let alertColor = "#333333"
+  let alertSrc = ""
+  let alertAlt = ""
+
+  if ( alert_icon == "beach_access" ) {
+  	let bgColor = "#EBFFF7"
```
jason-jackson, line 9: - Do we need these `let` in the if statements?
- Would it be cleaner to use a map?
- Would it be better to send generate this outside this file and pass it into it?

Schparky: something like this, Jason?
```
const alerts = {
  "beach_access": {
    {
      "bgColor": "#EBFFF7",
      "alertColor": "#028353",
...
    }
  },
}
...
<img src="cid:<%= alerts[alert_icon][alertSrc] %>" alt="<%= alerts[alert_icon][alertAlt] %>" style="width:100%;"/>
```

Schparky (approved):

jason-jackson:
Yes, @Schparky, though maybe we should also consider something more descriptive for names. `beach_access` is rather ambiguous and could cause confusion as it describes the icon, not the actual alert.

application/templates/mail/_alert.plush.html
```diff
@@ -1,35 +1,84 @@
 
-<div style="margin:16px 0px; padding:8px; border-radius: 8px; background: #EBEEF2;">
+<%
+	let alertStyles = {
+    	beach_access: {
+  			bgColor: "#EBFFF7",
+  			alertColor: "#028353",
+  			alertSrc: "beach_access",
```
jason-jackson, line 7: If wanted, you could omit this, as it's the same as alert_icon variable.

jason-jackson (approved):

PR #446, opened 2022-05-19
New values of ledger entry description and reference.

Baggerone:
This is to complete
https://trello.com/c/VDx5noff/545-revise-journal-entry-output

### DIFF
--- a/application/models/ledgerentry.go
+++ b/application/models/ledgerentry.go
@@ -78,6 +78,7 @@ type LedgerEntry struct {
 	UpdatedAt time.Time `db:"updated_at"`
 
 	Claim *Claim `belongs_to:"claims" validate:"-"`
+	Item  *Item  `belongs_to:"items" validate:"-"`
 }
 
 func (le *LedgerEntry) Create(tx *pop.Connection) error {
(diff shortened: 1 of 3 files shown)


jason-jackson (approved):

application/models/ledgerentry.go
```diff
@@ -196,25 +197,49 @@ func (le *LedgerEntry) Reconcile(ctx context.Context, now time.Time) error {
 	return nil
 }
 
-func (le *LedgerEntry) transactionDescription() string {
-	description := fmt.Sprintf("%s %s", le.RiskCategoryName, le.Type)
+// For household-type entries this returns `<entry.Type> <entry.Name>`.
+// For other entries this returns `<entry.Type> <entry.Name> (<accountable person name>)`,
+//   not including `<` and `>`
+func (le *LedgerEntry) transactionDescription(tx *pop.Connection) string {
```
Schparky, line 203: The idea was to not reference the database here, because database values can change over time, and we want the ledger to be static. If data is needed, it will have to be saved in the ledger entry at the time the record is created.

Schparky: I _think_ the `Name` property might work for what you're doing. Might need to double-check that it's correct in the real data set, if we want to be sure.

Baggerone: @Schparky, what if we make it check if there is already a description, in which case it returns the old description without making any changes?  If the old description is blank, then it gets new info.

Baggerone:
Try doing something without accessing the database

PR #447, opened 2022-05-20
Fix Ledger entry name and reference for journals

Baggerone:
https://docs.google.com/document/d/1E4lZGAxLFJ0LnNmHVTUGoIJ_adYIVVLNblb-BV1bXyc

https://trello.com/c/VDx5noff/545-revise-journal-entry-output

### DIFF
--- a/application/api/ledger.go
+++ b/application/api/ledger.go
@@ -70,8 +70,11 @@ type LedgerEntry struct {
 	AccountNumber    string          `json:"account_number"`
 	IncomeAccount    string          `json:"income_account"`
 
-	// name of accountable person if available, otherwise the policy name
-	Name   string   `json:"name"`
+	// name of accountable person if available
+	Name string `json:"name"`
+
+	PolicyName string `json:"policy_name"`
+
 	Amount Currency `json:"amount"`
 
 	// date added to ledger
--- a/application/migrations/20220520111000_add_fields_to_ledger_entries_table.up.fizz
+++ b/application/migrations/20220520111000_add_fields_to_ledger_entries_table.up.fizz
@@ -1,2 +1,19 @@
 add_column("ledger_entries", "policy_name", "string", {"default": ""})
-add_column("ledger_entries", "claim_payout_type", "string", {"default": ""})
+add_column("ledger_entries", "claim_payout_option", "string", {"default": ""})
+
+sql(`
+  UPDATE ledger_entries
+  SET policy_name = policies.name
+  FROM policies
+  WHERE policies.id = ledger_entries.policy_id
+`)
+
+sql(`
+  WITH options AS (
+    SELECT payout_option, claim_id FROM claim_items
+  )
+  UPDATE ledger_entries
+  SET claim_payout_option = options.payout_option
+  FROM options
+  WHERE options.claim_id = ledger_entries.claim_id
+`)
--- a/application/migrations/20220520111000_add_fields_to_ledger_entries_table.up.fizz
+++ b/application/migrations/20220520111000_add_fields_to_ledger_entries_table.up.fizz
@@ -1,0 +1,2 @@
+add_column("ledger_entries", "policy_name", "string", {"default": ""})
+add_column("ledger_entries", "claim_payout_type", "string", {"default": ""})
--- a/application/migrations/20220520111000_add_reference_to_ledger_entries_table.up.fizz
+++ b/application/migrations/20220520111000_add_reference_to_ledger_entries_table.up.fizz
@@ -1,1 +1,0 @@
-add_column("ledger_entries", "reference", "string", {"default": ""})
--- a/application/migrations/20220520111000_add_reference_to_ledger_entries_table.up.fizz
+++ b/application/migrations/20220520111000_add_reference_to_ledger_entries_table.up.fizz
@@ -1,0 +1,1 @@
+add_column("ledger_entries", "reference", "string", {"default": ""})
--- a/application/models/claim.go
+++ b/application/models/claim.go
@@ -926,11 +926,11 @@ func (c *Claim) CreateLedgerEntry(tx *pop.Connection) error {
 		item := c.ClaimItems[i].Item
 		item.LoadRiskCategory(tx, false)
 
-		le := NewLedgerEntry(c.Policy, &item, c)
+		le := NewLedgerEntry(tx, c.Policy, &item, c)
 		le.Type = LedgerEntryTypeClaim
 		le.Amount = c.TotalPayout
-		le.Description = le.getDescription(tx, &item, string(c.ClaimItems[i].PayoutOption), c.TotalPayout)
-		le.Reference = le.getReference(tx, &item)
+		le.Name = item.GetAccountablePersonName(tx).String()
+		le.ClaimPayoutOption = string(c.ClaimItems[i].PayoutOption)
 
 		le.RiskCategoryName = item.RiskCategory.Name
 		le.RiskCategoryCC = item.RiskCategory.CostCenter
--- a/application/models/claim.go
+++ b/application/models/claim.go
@@ -929,8 +929,8 @@ func (c *Claim) CreateLedgerEntry(tx *pop.Connection) error {
 		le := NewLedgerEntry(c.Policy, &item, c)
 		le.Type = LedgerEntryTypeClaim
 		le.Amount = c.TotalPayout
-		le.Name = le._onlyCreateName(tx, &item, string(c.ClaimItems[i].PayoutOption), c.TotalPayout)
-		le.Reference = le._onlyCreateReference(tx, &item)
+		le.Description = le.getDescription(tx, &item, string(c.ClaimItems[i].PayoutOption), c.TotalPayout)
+		le.Reference = le.getReference(tx, &item)
 
 		le.RiskCategoryName = item.RiskCategory.Name
 		le.RiskCategoryCC = item.RiskCategory.CostCenter
--- a/application/models/claim.go
+++ b/application/models/claim.go
@@ -925,8 +925,9 @@ func (c *Claim) CreateLedgerEntry(tx *pop.Connection) error {
 		c.ClaimItems[i].LoadItem(tx, false)
 		item := c.ClaimItems[i].Item
 		item.LoadRiskCategory(tx, false)
+		name := item.GetAccountablePersonName(tx).String()
 
-		le := NewLedgerEntry(tx, c.Policy, &item, c)
+		le := NewLedgerEntry(name, c.Policy, &item, c)
 		le.Type = LedgerEntryTypeClaim
 		le.Amount = c.TotalPayout
 
--- a/application/models/claim.go
+++ b/application/models/claim.go
@@ -929,8 +929,6 @@ func (c *Claim) CreateLedgerEntry(tx *pop.Connection) error {
 		le := NewLedgerEntry(tx, c.Policy, &item, c)
 		le.Type = LedgerEntryTypeClaim
 		le.Amount = c.TotalPayout
-		le.Name = item.GetAccountablePersonName(tx).String()
-		le.ClaimPayoutOption = string(c.ClaimItems[i].PayoutOption)
 
 		le.RiskCategoryName = item.RiskCategory.Name
 		le.RiskCategoryCC = item.RiskCategory.CostCenter
--- a/application/models/claim.go
+++ b/application/models/claim.go
@@ -926,15 +926,12 @@ func (c *Claim) CreateLedgerEntry(tx *pop.Connection) error {
 		item := c.ClaimItems[i].Item
 		item.LoadRiskCategory(tx, false)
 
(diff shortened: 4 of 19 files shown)


application/models/claim.go
```diff
@@ -926,15 +926,12 @@ func (c *Claim) CreateLedgerEntry(tx *pop.Connection) error {
...
-		name := item.GetAccountablePersonName(tx)
-
 		le := NewLedgerEntry(c.Policy, &item, c)
 		le.Type = LedgerEntryTypeClaim
 		le.Amount = c.TotalPayout
-		le.Name = name.String()
-		if le.Name == "" {
-			le.Name = c.Policy.Name
-		}
+		le.Name = le._onlyCreateName(tx, &item, string(c.ClaimItems[i].PayoutOption), c.TotalPayout)
```
jason-jackson, line 932: Should `_onlyCreateName` and `_onlyCreateReference just be a part of `NewLedgerEntry`? From what I can see, you don't need to pass anything else.

application/models/item.go
```diff
@@ -976,15 +977,11 @@ func (i *Item) CreateLedgerEntry(tx *pop.Connection, entryType LedgerEntryType,
...
-	name := i.GetAccountablePersonName(tx)
-
 	le := NewLedgerEntry(i.Policy, i, nil)
 	le.Type = entryType
 	le.Amount = -amount
-	le.Name = name.String()
-	if le.Name == "" {
-		le.Name = i.Policy.Name
-	}
+	le.Name = le._onlyCreateName(tx, i, "", amount)
```
jason-jackson, line 983: Should `_onlyCreateName` and `_onlyCreateReference just be a part of `NewLedgerEntry`? From what I can see, you don't need to pass anything else.

application/migrations/20220520111000_add_reference_to_ledger_entries_table.up.fizz
```diff
@@ -0,0 +1 @@
+add_column("ledger_entries", "reference", "string", {"default": ""})
```
Schparky, line 1: Do we need a SQL update to fill in this column for existing data?

application/models/ledgerentry.go
```diff
@@ -196,25 +226,70 @@ func (le *LedgerEntry) Reconcile(ctx context.Context, now time.Time) error {
 	return nil
 }
 
-func (le *LedgerEntry) transactionDescription() string {
-	description := fmt.Sprintf("%s %s", le.RiskCategoryName, le.Type)
+// _onlyCreateName should only be called when creating a ledgerEntry (not for subsequent usage)
+// For household-type entries this returns `<entry.Type.Description> · <Policy.Name>`.
+// For other entries this returns `<entry.Type.Description> · <Policy.Name> (<accountable person name>)`,
+//   not including `<` and `>`
+func (le *LedgerEntry) _onlyCreateName(tx *pop.Connection, item *Item, claimPayoutType string, amount api.Currency) string {
```
Schparky, line 233: I'm not too crazy about the leading underscore. Have you seen this done or recommended? Also, I'm not sure I understand the name. Is it named "onlyCreate" because it's only for calling on create? It seems confusing as a function name. Probably should stick with usual function naming paradigms and leave the "only create" warning to the doc comment. Try `getName` or `createNameField` or `getDescription`. Probably the latter, since it's more than just the name.

application/models/item.go
```diff
@@ -1027,6 +1024,9 @@ func (i *Item) GetAccountablePersonLocation(tx *pop.Connection) Location {
 // GetAccountablePerson gets the accountable person as a Person interface
 func (i *Item) GetAccountablePerson(tx *pop.Connection) Person {
 	var person Person
+	if i == nil {
```
Schparky, line 1027: I don't have a problem with this check, but it makes me wonder what situation arose that it became necessary.

application/models/ledgerentry.go
```diff
@@ -51,6 +51,34 @@ var ValidLedgerEntryTypes = map[LedgerEntryType]struct{}{
...
+		return "Coverage premium: Add"
+	case LedgerEntryTypeCoverageRenewal:
+		return "Coverage premium: Renew"
+	case LedgerEntryTypeCoverageRefund:
+		return "Coverage reimbursement: Remove"
+	case LedgerEntryTypeCoverageChange, LedgerEntryTypePolicyAdjustment:
+		if amount < 0 {
+			return "Coverage reimbursement: Reduce"
+		}
+		return "Coverage premium: Increase"
```
Schparky, line 66: I thought it was not possible to increase coverage.

application/models/ledgerentry.go
```diff
@@ -135,8 +165,8 @@ func (le *LedgerEntries) ToCsv(date time.Time) []byte {
 			sage.AppendToBatch(fin.Transaction{
 				Account:     domain.Env.ExpenseAccount,
 				Amount:      int(l.Amount),
-				Description: l.transactionDescription(),
-				Reference:   l.transactionReference(),
+				Description: l.Name,
```
Schparky, line 168: If the CSV column definition is changing from "description" to "name", it probably needs to stop using the `fin.Transaction` type and use a newly created one for this new CSV definition. I was pondering how we're getting away with redefining the Sage CSV format until I realized this isn't for import into Sage.

application/models/policy_test.go
```diff
@@ -225,6 +226,18 @@ func (ms *ModelSuite) TestPolicy_LoadMembers() {
 	ms.Len(policy.Members, 1)
 }
 
+func idsInOrder(id1, id2 uuid.UUID) string {
+	tmplt := `%s||%s`
```
Schparky, line 230:
```suggestion
	const template = `%s||%s`
```
I kept trying to read it as "temp less than" and it just didn't make sense!

application/models/ledgerentry.go
```diff
@@ -196,25 +226,70 @@ func (le *LedgerEntry) Reconcile(ctx context.Context, now time.Time) error {
...
+
+	description := le.Type.Description(claimPayoutType, amount)
+
+	if item == nil {
+		return description
+	}
+
+	item.LoadPolicy(tx, false)
+
+	description = fmt.Sprintf(`%s · %s`, description, item.Policy.Name)
```
Schparky, line 246: Is it safe to use non-ASCII characters in a CSV file? I'm particularly concerned Sage will choke on it.

application/models/ledgerentry.go
```diff
@@ -196,25 +226,70 @@ func (le *LedgerEntry) Reconcile(ctx context.Context, now time.Time) error {
...
-	return description
+	return fmt.Sprintf(`%s (%s)`, description, person)
 }
 
-// Merge HouseholdID and CostCenter into one column
-func (le *LedgerEntry) transactionReference() string {
+// _onlyCreateReference should only be called when creating a ledgerEntry (not for subsequent usage)
+// For household-type entries this returns `MC <entry.HouseholdID> / <accountable person name>`
+// For other entries this returns `<entry.EntityCode> <entry.AccountNumber><entry.CostCenter> / <Policy.Name>`,
+//   not including `<` and `>`.
```
Schparky, line 264: Just a thought, curly braces are often understood as interpolation symbols, so you could get away with using them without explanation, perhaps.

Schparky (changes requested):

application/models/ledgerentry.go
Baggerone: OK. (The leading underscore is more of a Python thing for methods other developers shouldn't depend on.)

application/models/ledgerentry.go
Baggerone: What should we use there instead?

application/models/ledgerentry.go
Baggerone: Do we know that will always be the case?  And even for now, if that's true, won't this still work, since the `amount` will be < 0?

application/models/claim.go
Baggerone: There would be other changes we would need to make to `NewLedgerEntry`. So, I think I'll leave it as I have it.

application/models/ledgerentry.go
Baggerone: Since curly braces are also understood as various other things, I think I'll leave it as is.

application/models/ledgerentry.go
Baggerone: (I've asked Alex on the Trello card.)

application/models/ledgerentry.go
Schparky: Yes, this code should be fine. Just checking my own understanding, as well as a question to be sure we're all on the same page.

application/models/ledgerentry.go
Schparky: I don't know, probably not. I'm confused about this today because my comments above don't appear to be correct. This is in `ToCsv` which is used to create the CSV for Sage. In that case, we can't redefine these columns at all, I think.

application/models/policy.go
```diff
@@ -641,7 +641,8 @@ func (p *Policy) CreateRenewalLedgerEntry(tx *pop.Connection, riskCategoryID uui
 	le := NewLedgerEntry(*p, nil, nil)
 	le.Type = LedgerEntryTypeCoverageRenewal
 	le.Amount = -amount
-	le.Name = p.Name
+	le.Name = le._onlyCreateName(tx, nil, "", amount)
```
Schparky, line 644: Rather than redefine the `Name` field, which would require updating previously-saved records, add new columns as necessary to save additional information needed later during report/export creation. That way the format can easily in the future, or new report definitions can be created without again mucking with historical data.

application/models/ledgerentry.go
Schparky: Thinking about other ways to address this so our future selves don't have to tiptoe around these functions. Move the calling code into the LedgerEntry definition so the "client" doesn't have to worry about implementation details. We have a "NewLedgerEntry" that creates a basic template. We could do one of two things: (1) create variations on that such as "NewLedgerEntryForPolicies" and "NewLedgerEntryForClaims", or (2) leave NewLedgerEntry as it is and add additional functions that are the "second step" of Ledger Entry creation. e.g. "CreatePolicyLedgerEntry", "CreateClaimLedgerEntry". These two helper functions (currently named `_onlyCreateName` and `_onlyCreateReference`) can be non-exported. Not that that prevents misuse, since everything is still in the same package, but it uses existing Go style to "document" that these are not for external use. Better yet, move some functionality into a sub-package. I don't think that's warranted at this time, but it might be good to start thinking in that way.

application/models/claim.go
jason-jackson: I'm fine with that, but amount is available from claim, which is passed to `NewLedgerEntry` as well.

application/models/ledgerentry.go
Baggerone: He said to use " / "

application/api/ledger.go
```diff
@@ -70,8 +70,11 @@ type LedgerEntry struct {
 	AccountNumber    string          `json:"account_number"`
 	IncomeAccount    string          `json:"income_account"`
 
-	// name of accountable person if available, otherwise the policy name
-	Name   string   `json:"name"`
+	// name of accountable person if available
+	Name string `json:"name"`
+
+	PolicyName string `json:"policy_name"`
```
Schparky, line 76: Is a UI change waiting for this? Just wondering if we need to time the releases around this feature.

application/migrations/20220520111000_add_fields_to_ledger_entries_table.up.fizz
```diff
@@ -0,0 +1,19 @@
...
+
+sql(`
+  UPDATE ledger_entries
+  SET policy_name = policies.name
+  FROM policies
+  WHERE policies.id = ledger_entries.policy_id
+`)
+
+sql(`
+  WITH options AS (
```
Schparky, line 12: Nice SQL work. I forgot about `WITH`!

application/models/item.go
```diff
@@ -976,15 +977,9 @@ func (i *Item) CreateLedgerEntry(tx *pop.Connection, entryType LedgerEntryType,
 	i.LoadRiskCategory(tx, false)
 	i.Policy.LoadEntityCode(tx, false)
 
-	name := i.GetAccountablePersonName(tx)
-
-	le := NewLedgerEntry(i.Policy, i, nil)
+	le := NewLedgerEntry(tx, i.Policy, i, nil)
```
Schparky, line 980: I'd prefer to see `name` passed as an argument to `NewLedgerEntry` rather than `tx`. It appears that it would work in every place it's used, but I may have missed something.

Schparky (approved):
Looks great. Just one suggestion.

application/api/ledger.go
Baggerone: Does the UI care about an unexpected extra field?

Schparky: No, but if it's expecting this field and it gets released first, that could be a problem.

PR #448, opened 2022-05-26
don't need getBuffaloContext since we're not using gqlgen

Schparky:


docker-compose.yml
```diff
@@ -81,8 +81,6 @@ services:
     image: postgres:11.6
     volumes:
       - ./db-init.sh:/docker-entrypoint-initdb.d/db-init.sh
-    ports:
-      - "5432:5432"
```
Baggerone, line 85: It might be nice to leave this in here as a reminder, instead of having to figure out what the default port is.

Baggerone (approved):

docker-compose.yml
Schparky: It conflicts when you run two apps side-by-side.

PR #449, opened 2022-06-02
More policy ledger table work

Baggerone:


application/models/ledgerreport.go
```diff
@@ -274,6 +266,34 @@ func newMonthlyPolicyLedgerReport(tx *pop.Connection, lEntries *LedgerEntries, l
 	return nil
 }
 
+func validateMonthYearForReport(month, year int) error {
```
Schparky, line 269: Good improvement in date validation!

application/models/ledgerreport.go
```diff
@@ -287,6 +307,57 @@ func newAnnualPolicyLedgerReport(tx *pop.Connection, lEntries *LedgerEntries, lR
...
+
+	if err := validateMonthYearForReport(month, year); err != nil {
+		return api.LedgerTable{}, err
+	}
+
+	var ledgerEntries LedgerEntries
+	startDate := time.Date(year, time.Month(month), 1, 0, 0, 0, 0, time.UTC)
+	endDate := startDate.AddDate(0, 1, -1)
+
+	q := "policy_id = ? AND date_entered >= ? AND date_entered <= ?"
```
Schparky, line 321:
```suggestion
	const q = "policy_id = ? AND date_entered BETWEEN ? AND ?"
```

application/models/policy.go
```diff
@@ -517,6 +517,18 @@ func (p *Policy) calculateAnnualPremium(tx *pop.Connection) api.Currency {
 	return premium
 }
 
+func (p *Policy) currentCoverage(tx *pop.Connection) api.Currency {
```
Schparky, line 520: Maybe this could be `currentCoverageTotal` to make it more clear it's a total coverage amount.

application/models/claim.go
```diff
@@ -908,7 +910,7 @@ func ClaimsWithRecentStatusChanges(tx *pop.Connection) (api.RecentClaims, error)
 }
 
 // CreateLedgerEntry does nothing if the TotalPayout is zero
-func (c *Claim) CreateLedgerEntry(tx *pop.Connection) error {
+func (c *Claim) CreateLedgerEntry(tx *pop.Connection, statusBefore, statusAfter string) error {
```
Schparky, line 913: `statusAfter` is always "Approved"

application/models/item.go
```diff
@@ -966,7 +975,8 @@ func (i *Item) SetAccountablePerson(tx *pop.Connection, id uuid.UUID) error {
 }
 
 // CreateLedgerEntry creates a charge of at least $1
-func (i *Item) CreateLedgerEntry(tx *pop.Connection, entryType LedgerEntryType, amount api.Currency) error {
+func (i *Item) CreateLedgerEntry(
+	tx *pop.Connection, entryType LedgerEntryType, amount api.Currency, statusBefore, statusAfter string) error {
```
Schparky, line 979: `statusAfter` is always "Approved" if `entryType` is "New" or "Changed", and  "Inactive" if `entryType` is "Refund".

application/migrations/20220602112000_add_statuses_to_ledger_entries_table.up.fizz
```diff
@@ -0,0 +1,3 @@
+add_column("ledger_entries", "status_before", "string", {"default": ""})
+add_column("ledger_entries", "status_after", "string", {"default": ""})
```
Schparky, line 2: Since `status_after` is easily derived from `type` do we need a new database column?

application/migrations/20220602112000_add_statuses_to_ledger_entries_table.up.fizz
```diff
@@ -0,0 +1,3 @@
+add_column("ledger_entries", "status_before", "string", {"default": ""})
```
Schparky, line 1: `status_before` can mostly be derived from `type` since the state flow chart dictates what previous states are allowed. The only one that's not entirely deterministic is whether an item was auto-approved or not. @merchako what is the value in adding the "previous status" to the ledger report?

Schparky (changes requested):
Some minor tweaks, and some design questions for @merchako.

application/migrations/20220602112000_add_statuses_to_ledger_entries_table.up.fizz
```diff
@@ -0,0 +1,3 @@
...
+add_column("ledger_entries", "status_after", "string", {"default": ""})
+
```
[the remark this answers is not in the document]
Baggerone: Dropping that idea of adding the columns to the db table

application/actions/policies.go
```diff
@@ -255,6 +265,68 @@ func policiesLedgerReportCreate(c buffalo.Context) error {
...
+func policiesLedgerTableView(c buffalo.Context) error {
+	policy := getReferencedPolicyFromCtx(c)
+
+	m := c.Param(MonthParam)
+	month, err := strconv.Atoi(m)
+	if err != nil {
+		appErr := api.AppError{
+			HttpStatus: http.StatusBadRequest,
+			Key:        api.ErrorInvalidDate,
+			Message:    fmt.Sprintf("%s, %s, must be a number", MonthParam, m),
```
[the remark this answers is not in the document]
Baggerone: Isn't enough that the swagger docs say they are required and an error is returned if they're invalid?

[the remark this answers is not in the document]
Schparky: I may have imagined that qs params should be optional. Couldn't find anything in writing to that effect.

Schparky (approved):

application/actions/policies.go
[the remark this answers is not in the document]
Baggerone: With our using the Authz middleware, I avoid using extra path parameters. And, apparently GET endpoints shouldn't expect a `body`.

PR #450, opened 2022-06-10
Fix missing ledger entry names

jason-jackson:
Load the data to fix the blank ledger entry name

### DIFF
--- a/application/models/ledgerentry.go
+++ b/application/models/ledgerentry.go
@@ -280,7 +280,7 @@ func (le *LedgerEntry) getReference() string {
 }
 
 func (le *LedgerEntry) getItemName(tx *pop.Connection) string {
-	le.LoadItem(tx)
+	le.LoadItem(tx, false)
 	if le.Item != nil {
 		return le.Item.Name
 	}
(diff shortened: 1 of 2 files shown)


application/models/ledgerentry.go
```diff
@@ -15,8 +15,10 @@ import (
 	"github.com/silinternational/cover-api/fin"
 )
 
-const accountSeparator = " / "
-const csvPolicyHeader = `"Amount","Description","Reference","Date Entered"` + "\n"
+const (
```
Schparky, line 18: Looks like you're running gofumpt. Nice.

application/models/ledgerentry.go
```diff
@@ -375,6 +378,15 @@ func (le *LedgerEntry) LoadClaim(tx *pop.Connection) {
 	}
 }
 
+// LoadItem - a simple wrapper method for loading the item
+func (le *LedgerEntry) LoadItem(tx *pop.Connection) {
```
Schparky, line 382: For consistency, I think we usually have a `reload bool` parameter.

application/models/ledgerentry.go
```diff
@@ -375,6 +378,15 @@ func (le *LedgerEntry) LoadClaim(tx *pop.Connection) {
...
+	if le.Item == nil {
```
Schparky, line 383:
```suggestion
	if le.Item == nil || reload {
```

Schparky (approved):

PR #452, opened 2022-06-15
Release

Schparky:
- set `ReadHeaderTimeout` to protect against Slowloris DDOS attack


hobbitronics (approved):
lgtm for what its worth

PR #456, opened 2022-09-02
Fix bug in cancelling coverage calculation

Baggerone:


application/models/item_test.go
```diff
@@ -890,40 +888,38 @@ func (ms *ModelSuite) TestItem_calculateCancellationCredit() {
...
+		oldPremium api.Currency
+		want       int
 	}{
 		{
-			name:        "no change",
-			coverage:    200000,
-			oldCoverage: 200000,
-			want:        0,
+			name:       "no change",
+			newPremium: 200000,
```
Schparky, line 902: Since these are premium amounts, not coverage amounts, maybe decrease the number to be more representative of typical values.

Schparky (approved):

PR #458, opened 2022-09-16
Release - refund calculation fix

Schparky:


codeship-steps.yml
```diff
@@ -5,10 +5,10 @@
     service: app
     command: ./codeship-test.sh
 
-  - name: apidocs
-    service: docs
-    tag: develop
-    command: ./build-push-api-docs.sh
+#  - name: apidocs
```
Baggerone, line 8: I don't remember if these are really supposed to be commented out.

Baggerone (approved):
One potential change to make.

codeship-steps.yml
Schparky: Wouldn't have expected this to be noticed, but the commit message was "disable api docs -- won't build but can't figure out why". I should have made that note on the PR.

PR #464, opened 2022-11-01
don't include zero-dollar transactions in the count

Schparky:
### Fixed
- Fix transaction count discrepancy due to inconsistency in handling of zero-dollar transactions.

### DIFF
--- a/application/models/ledgerreport.go
+++ b/application/models/ledgerreport.go
@@ -113,6 +113,10 @@ func (lr *LedgerReport) ConvertToAPI(tx *pop.Connection) api.LedgerReport {
 	transactionCount := len(lr.LedgerEntries)
 	isCleared := true
 	for _, e := range lr.LedgerEntries {
+		if e.Amount == 0 {
+			// TODO: consider whether we should even store ledger entries with a zero amount
+			transactionCount--
+		}
 		if !e.DateEntered.Valid {
 			isCleared = false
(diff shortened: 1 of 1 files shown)


application/models/ledgerreport.go
```diff
@@ -113,6 +113,10 @@ func (lr *LedgerReport) ConvertToAPI(tx *pop.Connection) api.LedgerReport {
 	transactionCount := len(lr.LedgerEntries)
 	isCleared := true
 	for _, e := range lr.LedgerEntries {
+		if e.Amount == 0 {
+			// TODO: consider whether we should even store ledger entries with a zero amount
```
mtompset, line 117: If zero is calculated as the amount for a particular transaction, then a zero dollar transaction proves the calculation was done. I don't know the underlying storage method or structure, but not storing it could create more problems, could it not?

mtompset (approved):
TODO comments in code tend to get stuck there forever. But the condition does look like it does what the description says.

application/models/ledgerreport.go
Schparky: That was my rationale for including them in the first place. I just don't want to cause more problems that I solve. And I agree about TODOs being a permanent fixture. This is more of an commentary, though, not really a problem to be solved. Maybe the "TODO" part was not necessary.

PR #468, opened 2022-11-16
Release 1.1.2 - dependency update

Schparky:
### Fixed
- Upgrade buffalo to 1.0.1
- Update all other dependencies except certmagic.


jason-jackson (approved):

Baggerone (approved):

Schparky:
> Isn't this a breaking change (like for the email templates)? Would a 1.2.0 be a better new release tag?

Dependency updates are usually a patch version. It's not a breaking change since the API didn't change.

PR #471, opened 2022-11-17
get client IP address from Cloudflare

Schparky:
### Fixed
- Get client IP address from Cloudflare instead of `RemoteAddr` since the latter will be the load balancer's IP address.

### DIFF
--- a/application/actions/actions.go
+++ b/application/actions/actions.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"net"
 	"net/http"
 
 	"github.com/gobuffalo/buffalo"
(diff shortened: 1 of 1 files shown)


application/actions/actions.go
```diff
@@ -129,3 +132,25 @@ func getUUIDFromParam(c buffalo.Context, param string) (uuid.UUID, error) {
...
+	req := c.Request()
+
+	// https://developers.cloudflare.com/fundamentals/get-started/reference/http-request-headers/#cf-connecting-ip
+	if cf := req.Header.Get("CF-Connecting-IP"); cf != "" {
+		return net.ParseIP(cf), nil
+	}
+
+	ip, _, err := net.SplitHostPort(req.RemoteAddr)
+	if err != nil {
+		return nil, fmt.Errorf("userip: %q is not IP:port, %w", req.RemoteAddr, err)
```
Baggerone, line 147: I'm just wondering if it would be good to go ahead and include the ip even if it's not a proper one (here and below) or maybe make the return string something like `<not a IP:port address>`. I'll let you decide. :-)

Baggerone (approved):

application/actions/actions.go
Schparky: The best solution would be to try and find the IP address from multiple places (RemoteAddr, X-Forwarding-For, etc.) but since we only care about our existing hosting environment I don't particularly want to deal with other scenarios right now.

Schparky: Thanks for the suggestion, though.

PR #472, opened 2022-12-13
Use Google Analytics 4 tracker

Baggerone:
I couldn't figure out how to get the branch name automatically included in the value of a codeship `environment` variable.


Baggerone:
> I think what you probably want to do is something like what the app-deployment-tracker-ga would do, and that was just have its code look for that environment variable automatically, without using having to configure our Codeship build to provide that

I would think it's better to allow ga_event_tracker to remain a general tool that doesn't know anything about the CI/CD service.

Baggerone:
> here's an example of using that in a command:

Thanks. That helped.

Schparky (approved):

codeship-services.yml
```diff
@@ -79,6 +79,23 @@ tracker:
...
+    # GA_EVENT_NAME is set by the codeship-steps command
+    # Add to the .env.encrypted file ... GA_API_SECRET
+
+ga4trackerFeature:
+  image: silintl/ga-event-tracker:1.0.1
+  encrypted_env_file: .env.encrypted
+  environment:
+    GA_MEASUREMENT_ID: G-NY4R8HM3Y3
+    GA_CLIENT_ID: cover-api
+    # Add to the .env.encrypted file ... GA_API_SECRET
```
forevermatt, line 97: This all looks identical to the `ga4tracker` entry just above it. Is there a difference (I could be missing it)? If not, you can just have the one, and have both of the codeship-steps entries use that single codeship-services entry.

codeship-steps.yml
```diff
@@ -61,6 +61,17 @@
...
   command: "true"
 
+- name: ga4track_deployment
+  service: ga4tracker
+  tag: ^(main|develop)
+  command: sh -c "export GA_EVENT_NAME=${CI_BRANCH}_CoverAPI && ./ga-event-tracker"
+
+# ${CI_BRANCH} can't be used here because it includes a slash
+- name: ga4track_feature_deployment
+  service: ga4trackerFeature
```
forevermatt, line 71: Suggested simplification:
```suggestion
  service: ga4tracker
```

forevermatt: Another question (perhaps you've already considered it) would be whether we want to bother tracking feature branches at all. 🤷

forevermatt (commented):
Suggested a way to simplify it, though I may have missed some important reason for the separate entries in codeship-services.yml.

codeship-services.yml
Baggerone: Good catch.

forevermatt (approved):
Looks good, thanks

PR #474, opened 2023-01-17
Release 1.3.0 

Baggerone:
Get newer version of node in the docker containers. 
Use Google Analytics 4 tracker 
get client IP address from Cloudflare 
generate a cert instead of using certmagic 
replace interface{} with any


application/cmd/app/cert.go
```diff
@@ -0,0 +1,79 @@
...
+		return fmt.Errorf("open cert file '%s': %w", certFile, err)
+	}
+	if err := pem.Encode(certOut, &pem.Block{Type: "CERTIFICATE", Bytes: derBytes}); err != nil {
+		return fmt.Errorf("write cert file '%s': %w", certFile, err)
+	}
+	if err := certOut.Close(); err != nil {
+		return fmt.Errorf("close cert file '%s': %w", certFile, err)
+	}
+
+	// #nosec G304 -- ok to use a filename from a variable since this is not exported code
```
Schparky, line 62:
```suggestion
	// #nosec G304 -- ok to use a filename from a variable since the filenames do not come from user data
```
I think I wrote this comment, but it doesn't seem quite right now.

Schparky (approved):

PR #475, opened 2023-01-24
add entityCodesCreate endpoint

hobbitronics:
add endpoint and tests


application/actions/app.go
```diff
@@ -37,7 +37,7 @@ import (
 	"github.com/gobuffalo/buffalo"
 	"github.com/gobuffalo/buffalo-pop/v3/pop/popmw"
 	contenttype "github.com/gobuffalo/mw-contenttype"
-	"github.com/gobuffalo/mw-i18n/v2"
+	i18n "github.com/gobuffalo/mw-i18n/v2"
```
hobbitronics, line 40: I don't remember changing this

application/actions/app.go
```diff
@@ -198,6 +198,7 @@ func App() *buffalo.App {
 		entityCodesGroup.GET("", entityCodesList)
 		entityCodesGroup.PUT(idRegex, entityCodesUpdate)
 		entityCodesGroup.GET(idRegex, entityCodesView)
+		entityCodesGroup.POST("/", entityCodesCreate)
```
Schparky, line 201:
```suggestion
		entityCodesGroup.POST("", entityCodesCreate)
```
Not sure what difference this makes, but line 198 has an empty string, so it's probably better to be consistent.

application/api/entitycodes.go
```diff
@@ -32,6 +32,9 @@ type EntityCode struct {
 
 // swagger:model
 type EntityCodeInput struct {
+	// Code is a unique three-letter identifier for an accounting entity
+	Code string `json:"code"`
```
Schparky, line 36: I think it would be better to create a new type `EntityCodeCreateInput` since this one is used by the update endpoint. As you have it, the update endpoint documentation will look like it's possible to change the `Code`. And future edits might actually make that possible when it must not be changed.

application/models/entitycode.go
```diff
@@ -32,6 +32,16 @@ func (ec *EntityCode) Create(tx *pop.Connection) error {
 	return create(tx, ec)
 }
 
+//create a new entity code from an api entity code
+func (ec *EntityCode) CreateFromAPI(tx *pop.Connection, input api.EntityCodeInput) error {
+	ec.Code = input.Code
```
Schparky, line 37:
```suggestion
	ec.ID = EntityCodeID(input.Code)
	ec.Code = input.Code
```
This is to create a V5 UUID instead of letting the ORM library, Pop, create a V4 UUID.

Schparky (changes requested):

application/actions/app.go
Baggerone: Maybe your IDE changed it?

application/actions/entitycodes_test.go
```diff
@@ -197,3 +197,64 @@ func (as *ActionSuite) Test_entityCodesView() {
...
+		wantError  api.ErrorKey
+	}{
+		{
+			name:       "must be authenticated",
+			actor:      models.User{},
+			wantStatus: http.StatusUnauthorized,
+			wantError:  api.ErrorNotAuthorized,
+		},
+		{
+			name:       "regular user cannot create endity codes",
```
Baggerone, line 220:
```suggestion
			name:       "regular user cannot create entity codes",
```

application/actions/entitycodes_test.go
```diff
@@ -197,3 +197,64 @@ func (as *ActionSuite) Test_entityCodesView() {
...
+		{
+			name:       "admin can create",
+			actor:      admin,
+			wantStatus: http.StatusOK,
+		},
+	}
+	for _, tt := range tests {
+		req := as.JSON("%s", entityCodesPath)
+		req.Headers["Authorization"] = fmt.Sprintf("Bearer %s", tt.actor.Email)
+		input := api.EntityCodeInput{
```
Baggerone, line 234: No big deal, but since this doesn't change, it could go before the `for` loop.

application/actions/entitycodes_test.go
```diff
@@ -197,3 +197,64 @@ func (as *ActionSuite) Test_entityCodesView() {
...
+
+		as.Equal(tt.wantStatus, res.Code, "incorrect status code returned: %d\n%s", res.Code, body)
+		if tt.wantError != "" {
+			var err api.AppError
+			as.NoError(as.decodeBody(body, &err), "response data is not as expected", body)
+			as.Equal(tt.wantError, err.Key, "error key is incorrect")
+		} else {
+			var code api.EntityCode
+			as.NoError(as.decodeBody(body, &code), "response data is not as expected %s", body)
+			as.NotNil(code.Active)
```
Baggerone, line 252: In a new PR, if you're interested, it may be nice to factor out the doubled up assertions into a single function that checks both of them.

application/actions/app.go
hobbitronics: that's what I'm thinking

Schparky:
Looks like you accidentally committed api-docs.html. Could you remove that file, please?

application/actions/app.go
Schparky: Either way, it's not a bad thing.

application/actions/entitycodes_test.go
```diff
@@ -197,3 +197,64 @@ func (as *ActionSuite) Test_entityCodesView() {
...
+			as.NotNil(code.IncomeAccount)
+			as.NotNil(code.ParentEntity)
+			as.Equal(input.Active, *code.Active)
+			as.Equal(input.IncomeAccount, *code.IncomeAccount)
+			as.Equal(input.ParentEntity, *code.ParentEntity)
+		}
+	}
+}
```
Schparky, line 260:
```suggestion
}

```
I don't like that VSCode doesn't add a linefeed at the end of the file. Well, to be accurate, I guess it's the red symbol in Github that I don't like.

Schparky (approved):

PR #476, opened 2023-01-25
remove the $1 minimum charge

Schparky:
### Changed
- Stop charging a full dollar for transactions that would be less than $1.

### DIFF
--- a/application/models/ledgerentry_test.go
+++ b/application/models/ledgerentry_test.go
@@ -591,7 +591,7 @@ func (ms *ModelSuite) Test_AdjustLedgerAmount() {
 			name:       "new coverage, under $1",
 			entryType:  LedgerEntryTypeNewCoverage,
 			amount:     50,
-			wantAmount: 100,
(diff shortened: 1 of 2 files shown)


application/models/ledgerentry_test.go
```diff
@@ -591,7 +591,7 @@ func (ms *ModelSuite) Test_AdjustLedgerAmount() {
 			name:       "new coverage, under $1",
 			entryType:  LedgerEntryTypeNewCoverage,
 			amount:     50,
-			wantAmount: 100,
+			wantAmount: 50,
```
hobbitronics, line 594: help me understand what the `wantAmount` is. I was thinking it should be the amount charged which would be a minimum of 100. Same for the one below.

hobbitronics (approved):
Approved as long as the test is correct. See comment

Schparky:
> Approved as long as the test is correct. See comment

The minimum only applies for credits now.

PR #477, opened 2023-01-25
Release 1.4.0

Schparky:
### Added
- New endpoint to allow admin users to create Entity Code records.
### Changed
- Stop charging a full dollar for transactions that would be less than $1.


Baggerone (approved):
Assuming you tested this manually on develop. :-)

PR #486, opened 2023-02-22
fix deactivated by claim message

Schparky:
### Fixed
- Added "approved claim" to "Deactivated by " status change message

### DIFF
--- a/application/models/item.go
+++ b/application/models/item.go
@@ -394,7 +394,7 @@ func (i *Item) cancelCoverageAfterClaim(tx *pop.Connection, reason string) error
 
 	i.CoverageEndDate = nulls.NewTime(now)
 	i.CoverageStatus = api.ItemCoverageStatusInactive
-	i.StatusChange = ItemStatusChangeInactivated + "approved claim"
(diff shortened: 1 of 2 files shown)


application/models/item.go
```diff
@@ -394,7 +394,7 @@ func (i *Item) cancelCoverageAfterClaim(tx *pop.Connection, reason string) error
 
 	i.CoverageEndDate = nulls.NewTime(now)
 	i.CoverageStatus = api.ItemCoverageStatusInactive
-	i.StatusChange = ItemStatusChangeInactivated
+	i.StatusChange = ItemStatusChangeInactivated + "approved claim"
```
Baggerone, line 397: This would give "Deactivated by approved claim".

I wonder if it would be more quickly understood if it were "Deactivated by having a claim approved"
So, this would be ... `+ "having a claim approved"`

Baggerone (approved):
Feel free to adopt my suggestion or ignore it.

application/models/item.go
Schparky: Thanks. Until you made that suggestion, it hadn't occurred to me that this message is displayed on the UI.

PR #487, opened 2023-02-28
audit to find items that were incorrectly renewed

Schparky:


application/actions/audits.go
```diff
@@ -0,0 +1,91 @@
...
+	if err := items.FindItemsIncorrectlyRenewed(tx, date); err != nil {
+		return result, err
+	}
+
+	if len(items) > MaxResultSize {
+		err := errors.New("too many rows in the result set")
+		return result, api.NewAppError(err, api.ErrorTooManyRows, api.CategoryInternal)
+	}
+
+	result.Items = items.ConvertToAPI(tx)
```
hobbitronics, line 88: Are the Items here still "filtered" by `FindItemsIncorrectlyRenewed`? I am finding it a little hard to understand how this works since `FindItemsIncorrectlyRenewed` seem to mutate them.

Schparky: Yes. You're correct that `FindItemsIncorrectlyRenewed` mutates `items`. It starts as an empty list. We structure it that way to organize the code around objects...not truly object oriented but close.

hobbitronics (approved):
I'm happy to approve this although I haven't tested it yet (just starting on a UI to do so). Feel free to wait on merging if you prefer though.

Schparky:
@Baggerone have you looked at this? Don't want to discredit @hobbitronics as a reviewer, but it would be good to have a seasoned Gopher review it as well.

application/models/item.go
```diff
@@ -1109,3 +1110,16 @@ func (i *Item) NewHistory(ctx context.Context, action string, fieldUpdate FieldU
...
 	}
 }
+
+// FindItemsIncorrectlyRenewed locates any items that were incorrectly renewed for another year of coverage. These are
+// identified as items that are marked as paid through the year but have an earlier coverage_end_date.
+func (i *Items) FindItemsIncorrectlyRenewed(tx *pop.Connection, date time.Time) error {
+	year := date.Year()
+	firstDayOfYear := time.Date(year, 1, 1, 0, 0, 0, 0, time.UTC)
+
+	err := tx.Where("paid_through_year >= ?", year).Where("coverage_end_date < ?", firstDayOfYear).All(i)
```
Baggerone, line 1120: Would it be worth including the maxResultSize in the query?
I guess this is how you would do it.
https://gobuffalo.io/documentation/database/querying/#join-query

If so, maybe make the page size one more than maxResultSize, in order to make it easy to test whether we should return an error.

Baggerone (approved):

application/models/item.go
Schparky: Not sure if that helps much, but an interesting idea.

PR #490, opened 2023-03-06
Release 1.5.0

Schparky:
### Added
- Added `can_be_deleted` and `can_be_updated` properties to API items
- Added new endpoint, `POST /audits`, to find items that were incorrectly renewed
- Added new endpoint, `POST /repairs`, to repair policy items that were processed during an annual renewal but should not have been.

### Included
- #487 
- #488 
- #489


application/actions/audits.go
```diff
@@ -0,0 +1,91 @@
...
+
+	var result api.AuditResult
+	var items models.Items
+	if err := items.FindItemsIncorrectlyRenewed(tx, date); err != nil {
+		return result, err
+	}
+
+	if len(items) > MaxResultSize {
+		err := errors.New("too many rows in the result set")
+		return result, api.NewAppError(err, api.ErrorTooManyRows, api.CategoryInternal)
```
jason-jackson, line 85: I think it would be better to limit the slice to `MaxResultSize` rather than throw an error.  For this instance it doesn't really matter as you can still see that there were a number of failures that need to be addressed, but if we apply it across the board, I think it would be better to `[:MaxResultSize]`.

application/actions/repairs.go
```diff
@@ -0,0 +1,87 @@
...
+func runRenewalRepair(c buffalo.Context, date time.Time) (api.RepairResult, error) {
+	tx := models.Tx(c)
+
+	var result api.RepairResult
+	var items models.Items
+	if err := items.RepairItemsIncorrectlyRenewed(c, date); err != nil {
+		return result, err
+	}
+
+	if len(items) > MaxResultSize {
```
jason-jackson, line 79: I think it would be better to limit it to first 1k results rather than return an error, but then again if there's more than 1k results, something is probably off.

jason-jackson, line 79: Also, just checking if this would rollback the above, so it doesn't throw an error while actually processing it.

jason-jackson (approved):
Left some comments, but nothing to hold it back or anything.

application/actions/repairs.go
Schparky: Right. As I mentioned on the other comment, I want to throw an error if we get too many records. It may mean the algorithm is wrong, not the data.

PR #496, opened 2023-03-14
process policy renewal in a background job

Schparky:
### Changed
- Moved the annual policy renewal process into a background job
- Added const identifiers for "development", "test", and "staging"
### Fixed
- Moved the initialization of listeners and jobs inside the `if app == nil` statement so they don't run twice. That constructor is only called twice in the compiled binary, so we didn't see this in development.

### DIFF
--- a/application/job/job.go
+++ b/application/job/job.go
@@ -116,7 +116,6 @@ func Submit(jobType string, args map[string]any) error {
 	if domain.Env.GoEnv == domain.EnvTest {
 		return nil
 	}
-	time.Sleep(time.Second)
 	job := worker.Job{
 		Queue:   "default",
 		Args:    args,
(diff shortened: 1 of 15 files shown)


application/job/job.go
```diff
@@ -110,44 +111,26 @@ func mainHandler(args worker.Args) error {
...
-func resubmitInactivateJob() {
-	// Run twice a day, in case it errors out
-	delay := time.Hour * 12
-
-	// uncomment this in development, if you want it to run more often for debugging
-	// delay = time.Second * 10
-
-	if err := SubmitDelayed(InactivateItems, delay, map[string]any{}); err != nil {
-		domain.ErrLogger.Printf("error resubmitting inactivateItemsHandler: " + err.Error())
+	time.Sleep(time.Second)
```
Baggerone, line 119: Just checking if this is needed now, e.g. was it just for troubleshooting.

Baggerone (approved):

application/job/job.go
Schparky: Good catch, thanks.

PR #500, opened 2023-03-15
Release 1.6.0 - Annual renewal process improvements

Schparky:
### Changed
- Don't allow low coverage amount on an item, as a backup for browser-based validation.
- Process policy renewal in a background job so the API doesn't block while processing.

### Added
- Added an API endpoint to query the annual renewal status.

### Internal
- Catch errors and panics in job handler.
- Refactor job handler.
- Get a newer version of ga-event-tracker.

### Included
- #494
- #495 
- #496 
- #497 
- #498 
- #499


Baggerone (approved):
Approved as long as you have been able to test that the new job handling code works properly. :-)

Schparky:
> Approved as long as you have been able to test that the new job handling code works properly. :-)

Yes, it works!

PR #501, opened 2023-03-23
use new log package

Schparky:
### Added
- New `log` package, using hooks for both Rollbar and Sentry (for now)
- Commit hash logging

### Removed
- Removed old loggers: `domain.Logger` and `domain.ErrLogger`
- Removed `domain.LogErrorMessage` function


Baggerone (approved):
Thanks for all your work on this.

PR #505, opened 2023-04-19
Fix user logging in Sentry

Schparky:
### Fixed
- Don't log user information on the global Sentry hub. This fixes the issue where user information appears on unrelated log entries.


Schparky:
This is an exact copy of the same file in ethnologue-cam

Baggerone (approved):

PR #511, opened 2023-06-14
CVR-9 return an HSTS header on every reply

briskt:
https://itse.youtrack.cloud/issue/CVR-9

### DIFF
--- a/application/domain/domain.go
+++ b/application/domain/domain.go
@@ -133,6 +133,7 @@ var Env struct {
 	ApiBaseURL                 string `required:"true" split_words:"true"`
 	AccessTokenLifetimeSeconds int    `default:"1166400" split_words:"true"` // 13.5 days
 	AppName                    string `default:"Cover" split_words:"true"`
+	HstsTtl                    int    `default:"3600" split_words:"true"` // default = 1 hour
 	LogLevel                   string `default:"debug" split_words:"true"`
 	AppNameLong                string `default:"Cover by SIL" split_words:"true"`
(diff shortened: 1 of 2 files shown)


application/domain/domain.go
```diff
@@ -133,6 +133,7 @@ var Env struct {
 	ApiBaseURL                 string `required:"true" split_words:"true"`
 	AccessTokenLifetimeSeconds int    `default:"1166400" split_words:"true"` // 13.5 days
 	AppName                    string `default:"Cover" split_words:"true"`
+	HstsTtl                    int    `default:"3600" split_words:"true"` // default = 1 hour
```
Baggerone, line 136: When I search Google for "what's a good default value for Strict-Transport-Security max-age", I see

`It is advisable to assign the max-age directive's value to be greater than 10368000 seconds (120 days) and ideally to 31536000 (one year).`

Baggerone (approved):

application/domain/domain.go
jason-jackson, line 136: What does `Ttl` stand for?  Would `HstsMaxAge` be clearer?

jason-jackson (approved):

application/domain/domain.go
briskt: I think the advice from @devon-sil was to start low and work up after you're more confident that nothing is going to break.

briskt: "Time To Live". I had thought someone called it that, but looking back at the docs there is no mention of that term in this context.

PR #513, opened 2023-06-29
allow the SAML key and certs to be given in simple Base64 format

briskt:
### Changed (non-breaking)
- Added flexibility to specify SAML certificates and private key in simple Base64 format as well as PEM format.


application/auth/saml/provider_test.go
```diff
@@ -199,35 +198,52 @@ func Test_getRsaPrivateKey(t *testing.T) {
...
+			wantErr:      true,
+			wantNumBytes: 0,
 		},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			got, err := pemToBase64(tt.pemStr)
+			got, err := decodeKey(tt.pemStr, tt.expectedType)
 			if (err != nil) != tt.wantErr {
 				t.Errorf("pemToBase64() error = %v, wantErr %v", err, tt.wantErr)
```
Baggerone, line 242: These error messages need to be updated, right? ("pemToBase64"  ---> "decideKey")

application/auth/saml/provider_test.go
```diff
@@ -199,35 +198,52 @@ func Test_getRsaPrivateKey(t *testing.T) {
...
+		},
+		{
+			name:         "valid PEM key",
+			pemStr:       ValidPrivateKey,
+			expectedType: "PRIVATE KEY",
+			wantErr:      false,
+			wantNumBytes: 1794,
+		},
+		{
+			name:         "wrong type",
```
Baggerone, line 231: It seems like there should be a test case for something that is already just a base64 string. (Or am I being blind to it :-)  )

application/auth/saml/provider.go
```diff
@@ -156,16 +155,11 @@ func getCertStore(cert string) (dsig.MemoryX509CertificateStore, error) {
...
+		return certStore, errors.New("a valid PEM or base64 encoded certificate is required")
 	}
 
-	cert, err := pemToBase64(cert)
-	if err != nil {
-		return certStore, err
-	}
-
-	certData, err := base64.StdEncoding.DecodeString(cert)
+	certData, err := decodeKey(cert, "CERTIFICATE")
```
Baggerone, line 162: If you aren't in too much of a hurry, you could make these type strings each a `const`.

Baggerone (approved):
Approved assuming you consider my remarks.

application/auth/saml/provider_test.go
briskt: Nope. I remember having a nagging feeling that I forgot something. I get that a lot.

PR #515, opened 2023-07-20
Feature/add netsuite export

jason-jackson:
### Added
- Added NetSuite ledger report

### Changed
- Ledger  report creation calls now take a `format` json field to specify whether Sage or NetSuite. Defaults to `Sage`.

https://itse.youtrack.cloud/issue/CVR-616/Create-new-export-for-NetSuite-Integration

### DIFF
--- a/application/fin/fin.go
+++ b/application/fin/fin.go
@@ -7,7 +7,10 @@ import (
 	"github.com/silinternational/cover-api/domain"
 )
 
-const ProviderTypeSage = "sage"
+const (
+	ProviderTypeNetSuite = "netsuite"
+	ProviderTypeSage     = "sage"
+)
 
 type Transaction struct {
 	Account     string
@@ -26,6 +29,13 @@ func NewBatch(providerType string, date time.Time) Provider {
 	batchDesc := fmt.Sprintf("%s %s JE", date.Format("January 2006"), domain.Env.AppName)
 
 	switch providerType {
+	case ProviderTypeNetSuite:
+		return &NetSuite{
+			Period:             getFiscalPeriod(int(date.Month())),
+			Year:               getFiscalYear(date),
+			JournalDescription: batchDesc,
+			Transactions:       nil,
+		}
 	case ProviderTypeSage:
 		return &Sage{
 			Period:             getFiscalPeriod(int(date.Month())),
--- a/application/fin/fin.go
+++ b/application/fin/fin.go
@@ -43,6 +43,7 @@ type Transaction struct {
 type Provider interface {
 	AppendToBatch(Transaction)
 	BatchToCSV() []byte
+	
 	getDescription(Transaction) string
 	getReference(Transaction) string
 }
--- a/application/fin/fin.go
+++ b/application/fin/fin.go
@@ -4,25 +4,47 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/silinternational/cover-api/api"
 	"github.com/silinternational/cover-api/domain"
 )
 
 const (
 	ProviderTypeNetSuite = "netsuite"
+	ProviderTypePolicy   = "policy"
 	ProviderTypeSage     = "sage"
 )
 
+type (
+	Transactions      []Transaction
+	TransactionBlocks map[string]Transactions // keyed by account
+)
+
 type Transaction struct {
+	EntityCode        string
+	RiskCategoryName  string
+	RiskCategoryCC    string // Risk Category Cost Center
+	Type              string
+	PolicyType        api.PolicyType
+	HouseholdID       string
+	CostCenter        string
+	AccountNumber     string
+	IncomeAccount     string
+	Name              string
+	PolicyName        string
+	ClaimPayoutOption string
+
 	Account     string
 	Amount      int
 	Description string
-	Reference   string
+	Reference   *string // Override the reference if given
 	Date        time.Time
 }
 
 type Provider interface {
 	AppendToBatch(Transaction)
 	BatchToCSV() []byte
+	getDescription(Transaction) string
+	getReference(Transaction) string
 }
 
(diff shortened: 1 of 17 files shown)


application/models/ledgerentry.go
```diff
@@ -229,10 +229,32 @@ func (le *LedgerEntry) Reconcile(ctx context.Context, now time.Time) error {
...
 }
 
+// getAccount returns text that is based on which report is being run and
+// LedgerEntry policy type (not including `<` and `>`)
+//
+// - For Sage reports, it returns the expense account
+//
+// - For household-type entries this returns `<entry.HouseholdID>`
+//
+// - For other entries this returns `<entry.EntityCode>`
```
briskt, line 239: This doesn't seem to match the printf at the end of the function. Actually, this comment block doesn't really add much to the code, which is pretty easy to read.

application/models/ledgerentry.go
```diff
@@ -229,10 +229,32 @@ func (le *LedgerEntry) Reconcile(ctx context.Context, now time.Time) error {
...
+	return fmt.Sprintf("%s %s%s / %s",
+		le.EntityCode, le.AccountNumber, le.CostCenter, le.PolicyName)
+}
+
 // getDescription returns text that is base on other fields of the LedgerEntry
 // For household-type entries this returns `<entry.Type.Description> / <Policy.Name>`.
 // For other entries this returns `<entry.Type.Description> / <Policy.Name> (<accountable person name>)`,
-//   not including `<` and `>`
+//
+//	not including `<` and `>`
```
briskt, line 257: go fmt has changed since this code was written. The indent triggers a newline, which is not really wanted here. So you might need to remove the space before "not including".

application/models/ledgerentry.go
```diff
@@ -257,8 +279,8 @@ func (le *LedgerEntry) getDescription() string {
 // getReference returns text that is base on other fields of the LedgerEntry
 // For household-type entries this returns `MC <entry.HouseholdID> / <accountable person name>`
 // For other entries this returns `<entry.EntityCode> <entry.AccountNumber><entry.CostCenter> / <Policy.Name>`,
-//   not including `<` and `>`.
-func (le *LedgerEntry) getReference() string {
+// not including `<` and `>`.
+func (le *LedgerEntry) getReference(format string) string {
```
briskt, line 283: why is this new argument added, but not used?

briskt (commented):
a few comments added, but I want to review it a bit more. Would like to see if it's possible to keep the format isolated to the fin package and not leak that out into the models package.

jason-jackson:
> a few comments added, but I want to review it a bit more. Would like to see if it's possible to keep the format isolated to the fin package and not leak that out into the models package.

Yeah, I was thinking the same, probably doable by adding more fields to Transaction.

jason-jackson:
@briskt refactored it some more, moving more to fin package. The block creation could almost be moved too, but I'd need to think about how to do the balance descriptions for this (Line 191 of models/ledgerentry.go)
```			Description: ledgerEntries[0].balanceDescription(),```

File: application/fin/netsuite.go [hunk 4]
```diff
@@ -0,0 +1,101 @@
...
+	return []byte(str)
+}
+
+func (n *NetSuite) transactionRow(rowNumber int) []byte {
+	t := n.Transactions[rowNumber]
+	str := fmt.Sprintf(
+		netSuiteTransactionRowTemplate,
+		n.getAccount(t),
+		api.Currency(-t.Amount).String(),
+		fmt.Sprintf("%.60s", n.getDescription(t)), // TODO does NetSuite have a 60 char field limit?
```
briskt, line 96: Have you asked someone about this TODO question? Chances are good that there is a limit and that it's different than 60.

application/fin/netsuite.go
```diff
@@ -0,0 +1,101 @@
...
+	buf.Write(n.summaryRow())
+	for i := range n.Transactions {
+		buf.Write(n.transactionRow(i))
+	}
+
+	return buf.Bytes()
+}
+
+func (n *NetSuite) getAccount(t Transaction) string {
+	if t.Account != "" {
```
briskt, line 44: I don't see this part in the Youtrack issue. From what I see the account column is either the household ID or the entity code, but never the account number.

application/fin/fin.go
```diff
@@ -4,28 +4,62 @@ import (
...
 	Description string
-	Reference   string
+	Reference   *string // Override the reference if given
 	Date        time.Time
 }
 
 type Provider interface {
 	AppendToBatch(Transaction)
 	BatchToCSV() []byte
+	getDescription(Transaction) string
```
briskt, line 46: `getDescription` is the same and very simple on all three providers, so I wonder if it would be better to eliminate this method.

briskt (approved):
I'm going to approve it, though I have a few suggestions and questions. All-in-all it looks very good!

application/fin/netsuite.go
jason-jackson: I have this in here so I can override it for category summary lines.

application/fin/fin.go
```diff
@@ -4,28 +4,62 @@ import (
...
+	Name              string
+	PolicyName        string
+	ClaimPayoutOption string
+
 	Account     string
-	Amount      int
+	Amount      api.Currency
 	Description string
-	Reference   string
+	Reference   *string // Override the reference if given
```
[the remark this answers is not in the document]
jason-jackson: There are times I want a blank reference field, hence the nil check. I could base that on other fields given, instead though. Or I could switch to nulls.String too.

application/fin/fin.go
jason-jackson: I could, we could always add it again later if we need to change it for a specific provider.

briskt (approved):

PR #516, opened 2023-07-24
Split household and team claims in report

jason-jackson:
### Changed

- Split household and team claims in report

I think I'd like to add a couple more tests, but starting a PR for ideas/suggestions.


application/models/ledgerentry.go
```diff
@@ -201,6 +202,11 @@ func (le *LedgerEntries) MakeBlocks() TransactionBlocks {
 	blocks := TransactionBlocks{}
 	for _, e := range *le {
 		key := e.IncomeAccount + e.RiskCategoryCC
+
+		// Split claims up by Policy Type, using space to trim later to keep same account number
+		if e.Type.IsClaim() && e.PolicyType == api.PolicyTypeHousehold {
+			key += " "
```
briskt, line 208: It feels a bit like hocus-pocus to me. What would you think of adding the account as a field so that the key can be throwaway? Then you could append something more meaningful like " Household". Actually `AccountNumber` is already a field. I can't recall whether that's the same thing.

briskt: A bit better. Can we use the AccountNumber field instead of the map key so the prefix doesn't have to be removed?

jason-jackson: Not sure if that is feasible, it's used for team policies.

briskt (approved):

PR #519, opened 2023-07-30
Create zip for Netsuite, create 2 reports on call

jason-jackson:
Alternate PR to #518, I wanted to see if this looked any cleaner. 

### Changed
- Create Ledger report makes 2 reports, both Sage CSV and NetSuite Zip file
- Added more information to Readme about `cover.local`

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,11 @@ The only system requirement for running this application is Docker. Once the sou
 to do to get it running is:
 
 1. Copy `local-example.env` to `local.env` and update values as described in the file. Secrets may be provided by another team member via Signal or other secure communication tool.
-1. Add `127.0.0.1 minio` to `/etc/hosts` (or equivalent for your OS)
+1. Add the following to `/etc/hosts` (or equivalent for your OS)
+   ```
+   127.0.0.1 minio
+   127.0.0.1 cover.local
+   ```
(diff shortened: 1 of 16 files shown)


application/actions/ledger.go
```diff
@@ -114,6 +111,31 @@ func ledgerReportCreate(c buffalo.Context) error {
 		return reportError(c, err)
 	}
 
+	// Create NetSuite report
```
jason-jackson, line 114: I'm debating adding a `NewLedgerReport` function to LedgerEntries that would move the bulk of this out of here.

README.md
jason-jackson, line 1: Looks like formatter kicked in, I could revert and try again if needed.

jason-jackson:
Pros:

- Don't need to modify database
- Don't need to modify UI (potentially, need to check if UI cares about return of Ledger Report Create)
- Easy to switch if we change providers later on

briskt (approved):
This does look pretty good. You may want to add a column to the database to store the report format. The content type (csv/zip) can help, but it's not quite the same. Not sure whether it would go on the API response or not, I guess that's up to the UI/UX people.

jason-jackson:
The file name also has the report format, so that could be used as well.

PR #521, opened 2023-08-08
Feature/additional netsuite export requests

jason-jackson:
### Changed

- Add report date to csv filenames in NetSuite report
- Removed summary rows from NetSuite report
- Removed RECTYPE, SCURNDEC, SRCELDGR, SRCETYPE columns from NetSuite report
- Added TRANSID column to NetSuite report in the format <reportYear><reportMonth><annualFlag><incrementingNumber>

### DIFF
--- a/application/fin/fin.go
+++ b/application/fin/fin.go
@@ -47,17 +47,22 @@ type Report interface {
 	getReference(Transaction) string
 }
 
-func NewBatch(reportFormat string, date time.Time) Report {
+func NewBatch(reportFormat, reportType string, date time.Time) Report {
 	batchDesc := fmt.Sprintf("%s %s JE", date.Format("January 2006"), domain.Env.AppName)
 
 	switch reportFormat {
 	case ReportFormatNetSuite:
+		annual := 0
+		if reportType == "Annual" {
+			annual = 1
+		}
 		return &NetSuite{
-			Date:               date,
 			Period:             getFiscalPeriod(int(date.Month())),
 			Year:               getFiscalYear(date),
 			JournalDescription: batchDesc,
 			TransactionBlocks:  make(TransactionBlocks),
+			date:               date,
+			annual:             annual,
 		}
 	case ReportFormatPolicy:
 		return &Policy{}
--- a/application/fin/fin.go
+++ b/application/fin/fin.go
@@ -53,6 +53,7 @@ func NewBatch(reportFormat string, date time.Time) Report {
 	switch reportFormat {
 	case ReportFormatNetSuite:
 		return &NetSuite{
+			Date:               date,
 			Period:             getFiscalPeriod(int(date.Month())),
 			Year:               getFiscalYear(date),
 			JournalDescription: batchDesc,
--- a/application/fin/netsuite.go
+++ b/application/fin/netsuite.go
@@ -19,11 +19,13 @@ const (
 var netsuiteRowNum int
 
 type NetSuite struct {
-	Date               time.Time
 	Period             int
 	Year               int
 	JournalDescription string
 	TransactionBlocks  TransactionBlocks
+
+	date   time.Time
+	annual int
 }
 
 func (n *NetSuite) AppendToBatch(block string, t Transaction) {
@@ -40,7 +42,7 @@ func (n *NetSuite) RenderBatch() ([]byte, string) {
 	w := zip.NewWriter(buff)
 
 	for blockName, block := range n.TransactionBlocks {
-		f, err := w.Create(fmt.Sprintf("%s_%s.csv", blockName, n.Date.Format(domain.DateFormat)))
+		f, err := w.Create(fmt.Sprintf("%s_%s.csv", blockName, n.date.Format(domain.DateFormat)))
 		if err != nil {
 			log.Fatal(err)
 		}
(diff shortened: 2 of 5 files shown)


application/fin/netsuite.go
```diff
@@ -5,24 +5,27 @@ import (
...
+var netsuiteRowNum int
+
 type NetSuite struct {
 	Period             int
 	Year               int
 	JournalDescription string
 	TransactionBlocks  TransactionBlocks
+
+	date   time.Time
+	annual int
```
briskt, line 28: This feels like it should be a boolean. Otherwise, it is not clear what value it should hold.
```suggestion
	annual bool
```

application/fin/fin.go
```diff
@@ -47,16 +47,22 @@ type Report interface {
...
-func NewBatch(reportFormat string, date time.Time) Report {
+func NewBatch(reportFormat, reportType string, date time.Time) Report {
 	batchDesc := fmt.Sprintf("%s %s JE", date.Format("January 2006"), domain.Env.AppName)
 
 	switch reportFormat {
 	case ReportFormatNetSuite:
+		annual := 0
+		if reportType == "Annual" {
+			annual = 1
+		}
```
briskt, line 58: I think a boolean is better. Also, "Annual" should be a const since it is a critical value in a comparison. You could even change `reportType string` to `isAnnual bool`. If you think you need more flexibility, make a `map` that converts `reportType` to a single-character type code that gets inserted into the transaction ID (row ID?). As it is, a future change that assigns a value > 9 would cause the transaction ID to increase in length. That seems to be a bad thing since it appears to be a conglomeration of a bunch of information packed into a short string with no delimiters.
```suggestion
		annual := false
		if reportType == "Annual" {
			annual = true
		}
```

application/fin/netsuite.go
```diff
@@ -5,24 +5,27 @@ import (
...
 	"fmt"
 	"log"
+	"time"
 
 	"github.com/silinternational/cover-api/api"
 	"github.com/silinternational/cover-api/domain"
 )
 
 const (
-	netSuiteHeader1 = `"RECTYPE","BATCHID", "BTCHENTRY","ORIGCOMP","SRCELEDGER","SRCETYPE","FSCSYR","FSCSPERD","SWEDIT","JRNLDESC","REVPERD","ERRBATCH","ERRENTRY","DETAILCNT","PROCESSCMD"` + "\n"
```
briskt, line 14: It's good to see this stuff go away. It always made my skin crawl to call this a CSV.

application/fin/netsuite.go
```diff
@@ -5,24 +5,27 @@ import (
...
-	netSuiteHeader2 = `"RECTYPE","ACCTID","TRANSAMT","SCURNDEC","TRANSDESC","TRANSREF","TRANSDATE","SRCELDGR","SRCETYPE"` + "\n"
-
-	netSuiteSummaryRowTemplate     = `"1","000000","00001","","GL","JE","%d","%02d",0,"%s","00",0,0,0,2` + "\n"
-	netSuiteTransactionRowTemplate = `"2","%s",%s,"2","%s","%s",%s,"GL","JE"` + "\n"
+	netSuiteHeader                 = `"TRANSID","ACCTID","TRANSAMT","TRANSDESC","TRANSREF","TRANSDATE"` + "\n"
+	netSuiteTransactionRowTemplate = `%s,"%s",%s,"%s","%s",%s` + "\n"
 )
 
+var netsuiteRowNum int
```
briskt, line 19: Uh-oh, using a global is often a bad thing. In this case, I am sure it will be a bug because it is never reset to 0 until the next time the container restarts. Please move it into a function parameter or struct field.

application/fin/netsuite.go
```diff
@@ -107,14 +108,15 @@ func (n *NetSuite) getReference(t Transaction) string {
 	return fmt.Sprintf("%s / %s", ref, t.PolicyName)
 }
 
-func (n *NetSuite) summaryRow() []byte {
-	str := fmt.Sprintf(netSuiteSummaryRowTemplate, n.Year, n.Period, n.JournalDescription)
-	return []byte(str)
+func (n *NetSuite) getRowID() string {
+	return fmt.Sprintf("%d%02d%d%06d", n.Year, n.Period, n.annual, netsuiteRowNum)
```
briskt, line 112: Is this a NetSuite requirement or something that Andy came up with? Why not put each of these values in its own column?

application/fin/netsuite_test.go
```diff
@@ -51,10 +52,8 @@ func TestNetSuite_Export(t *testing.T) {
 		TransactionBlocks:  TransactionBlocks{"": Transactions{t1}, "bar": Transactions{t2}},
 	}
 
-	summaryRow := fmt.Sprintf(`"1","000000","00001","","GL","JE","%d","%02d",0,"%s","00",0,0,0,2`+"\n",
-		n.Year, n.Period, n.JournalDescription)
-
 	transaction1Row := fmt.Sprintf(netSuiteTransactionRowTemplate,
+		fmt.Sprintf("%d%02d0%06d", n.Year, n.Period, 1),
```
briskt, line 56: What do you think of this? I think it feels more like test code this way.
```suggestion
		"2009090000001",
```

application/fin/netsuite_test.go
```diff
@@ -63,6 +62,7 @@ func TestNetSuite_Export(t *testing.T) {
 	)
 
 	transaction2Row := fmt.Sprintf(netSuiteTransactionRowTemplate,
+		fmt.Sprintf("%d%02d0%06d", n.Year, n.Period, 2),
```
briskt, line 65:
```suggestion
		"2009090000002",
```

briskt (changes requested):
I have a few suggestions, only one of which is a must-change.

briskt (approved):

PR #522, opened 2023-08-15
split out cost center and account from transref

jason-jackson:
### Changed

- Split out cost center and account from TRANSREF column

Issue: https://itse.youtrack.cloud/issue/CVR-640

### DIFF
--- a/application/fin/netsuite.go
+++ b/application/fin/netsuite.go
@@ -12,8 +12,8 @@ import (
 )
 
 const (
-	netSuiteHeader                 = `"TRANSID","ACCTID","TRANSAMT","TRANSDESC","TRANSREF","TRANSDATE"` + "\n"
-	netSuiteTransactionRowTemplate = `%d,"%s",%s,"%s","%s",%s` + "\n"
+	netSuiteHeader                 = `"TRANSID","EntityCode","TRANSAMT","TRANSDESC", "Account", "SuiteKey","TRANSREF","TRANSDATE"` + "\n"
+	netSuiteTransactionRowTemplate = `%d,"%s",%s,"%s","%s","%s","%s",%s` + "\n"
 )
 
 type NetSuite struct {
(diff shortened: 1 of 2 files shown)


application/fin/netsuite.go
```diff
@@ -12,8 +12,8 @@ import (
 )
 
 const (
-	netSuiteHeader                 = `"TRANSID","ACCTID","TRANSAMT","TRANSDESC","TRANSREF","TRANSDATE"` + "\n"
-	netSuiteTransactionRowTemplate = `%d,"%s",%s,"%s","%s",%s` + "\n"
+	netSuiteHeader                 = `"TRANSID","EntityCode","TRANSAMT","TRANSDESC", "Account", "SuiteKey","TRANSREF","TRANSDATE"` + "\n"
```
briskt, line 15: I don't like that this column header is "EntityCode" since that label is used for the three-character organizational unit identifier both internal to the code and in the user interface.
In the linked YT ticket, I believe that label was just a suggestion: "Maybe “EntityCode”?". Would `"AccountReference"` work? If time allows, it would probably be better to establish proper naming with our customer. If "EntityCode" is really the most accurate label for this, then we should think about renaming every one of our references to that term. But I really hope that we didn't get it wrong.

briskt (changes requested):

application/fin/netsuite.go
jason-jackson: From Andy:
> We should probably change the label of the ACCTID field as well since we’re adding a new column with the actual GL Account.    The ACCTID field contains Entity Codes and Ministry IDs.   Not sure what’s an appropriate name.   Maybe “EntityCode”?

So there's definitely some flexibility to naming this, I'll see what he thinks about AccountReference

jason-jackson: This field is either Household ID or EntityCode, so `EntityCode` is ok with me, but the main purpose of this PR is to split up multipurpose columns, so maybe it might be worth splitting up ACCTID as well?

 `AccountReference` seems a bit similar to `TRANSREF`, but I suppose it might be fine.

I also don't mind just leaving it `ACCTID`, but he mentioned it might be good to change it, so it probably should be.

briskt: > so maybe it might be worth splitting up ACCTID as well?

I like that idea, but can you do that without permission?

briskt (approved):

application/fin/netsuite.go
jason-jackson: Nope, which was why I asked him. I didn't think it was entirely viable.

> >  Correct me if I'm wrong, but I don't think `ACCTID` is a field we can split up like TRANSREF. Would using `EntityAccount` instead of `Account` work for the new column instead of changing the `ACCTID` header?
> 
> You’re right. Splitting up the data in the ACCTID column would take a lot more work.  I’m fine with EntityAccount.

briskt (approved):

PR #523, opened 2023-08-30
create a service user for logging activity by scheduled jobs and events

briskt:
### Added
- Insert a "service user" into the `users` table with a predefined ID.

### Changed
- Use the ID of the service user in the policy history record when canceling coverage after a claim.
- Identify the service user as the "actor" during all job processing.

Issue: [CVR-636](https://itse.youtrack.cloud/issue/CVR-636)


application/models/user.go
```diff
@@ -197,6 +198,15 @@ func GetDefaultSteward(tx *pop.Connection) User {
 	return u
 }
 
+// GetServiceUser returns the Service User record
+func GetServiceUser(tx *pop.Connection) User {
+	u := User{}
+	if err := tx.Find(&u, ServiceUserID); err != nil {
+		panic("error finding service user" + err.Error())
```
jason-jackson, line 205:
```suggestion
		panic("error finding service user: " + err.Error())
```

jason-jackson (approved):
Should you use logging fatal instead of panic?

PR #528, opened 2023-08-31
include the entity code `Active` flag for non-admin users

briskt:
### Fixed
- Include the entity code `Active` flag for non-admin users to fix the searchable select on the `/policies/new` UI page.

Issue: [CVR-604](https://itse.youtrack.cloud/issue/CVR-604)


jason-jackson (approved):

application/actions/entitycodes_test.go
```diff
@@ -67,11 +67,9 @@ func (as *ActionSuite) Test_entityCodesList() {
```
jason-jackson, line 67: Would still be good to check for this.

```suggestion
			as.NotNil(codes[0].Active)
```

PR #532, opened 2023-09-05
Add fields for vehicle coverage

jason-jackson:
### Changed

- Add year and monthly premium fields for vehicle coverage
- Change docker environment variables to be of type string instead of bool


application/models/item.go
```diff
@@ -841,8 +842,14 @@ func (i *Item) CalculateProratedPremium(t time.Time) api.Currency {
 	return api.Currency(p)
 }
 
+// CalculateMonthlyPremium returns the rounded product of the item's CoverageAmount and the PremiumFactor
+func (i *Item) CalculateMonthlyPremium() api.Currency {
+	p := int(math.Round(float64(i.CoverageAmount) * domain.Env.VehiclePremiumFactor))
```
briskt, line 847: This function doesn't seem to be used yet, so I'm not sure if it's named incorrectly, or the wrong premium factor is used.

application/api/items.go
```diff
@@ -100,6 +103,9 @@ type Item struct {
 	// estimated annual premium prorated from now to the end of the year (0.01 USD)
 	ProratedAnnualPremium Currency `json:"prorated_annual_premium"`
 
+	// monthly premium (0.01 USD)
+	MonthlyPremium Currency `json:"monthly_premium"`
```
briskt, line 107: This isn't used anywhere. Did you intend to connect it to `CalculateMonthlyPremium`?

application/grifts/db.go
```diff
@@ -301,6 +301,8 @@ INSERT INTO "item_categories" ("id", "risk_category_id", "name", "help_text",
     'Enabled',	300000,	'2021-08-27 19:46:28',	'2021-08-27 19:46:28',	8, false),
 ('036e5315-18ca-4404-8435-72a695f2c9a7',	'3be38915-7092-44f2-90ef-26f48214b34f',	'Travel and recreation',	'Includes suitcases, travel bags, cycling, skating, sports. No motorized vehicles.',
     'Enabled',	300000,	'2021-08-27 19:46:28',	'2021-08-27 19:46:28',	9, true);
+('0619a0ba-785e-428d-858c-96d3bd56929a',	'3be38915-7092-44f2-90ef-26f48214b34f',	'Cars and Heavy Vehicles',	'Coverage for Cars and Heavy Vehicles does not include liability and is not intended to fulfill local regulations or compete with local insurance offerings.',
+    'Enabled',	8000000,	'2021-08-27 19:46:28',	'2021-08-27 19:46:28',	9, true);
```
briskt, line 305:
```suggestion
    'Enabled',	8000000,	'2021-08-27 19:46:28',	'2021-08-27 19:46:28',	NULL, true);
```

briskt (changes requested):

application/models/item.go
jason-jackson: I've added it to ConvertToAPI now. I probably should make it nullable as I don't think we need to return monthly premiums on non-vehicles. trying to decide the best way to do that.

application/migrations/20230829220027_add_vehicle_coverage_fields.down.fizz
```diff
@@ -0,0 +1 @@
+# drop_column("items", "year")
```
briskt, line 1: There's a `#` that makes this a comment, disabling the `drop_column`.
```suggestion
drop_column("items", "year")
```

application/migrations/20230829220027_add_vehicle_coverage_fields.up.fizz
```diff
@@ -0,0 +1 @@
+# add_column("items", "year", "int", {"null": true})
```
briskt, line 1: There's a `#` that makes this a comment, disabling the `add_column`.
```suggestion
add_column("items", "year", "int", {"null": true})
```

briskt (changes requested):
Found the issue with your migration

application/models/item.go
```diff
@@ -1187,3 +1195,8 @@ func CountItemsToRenew(tx *pop.Connection, year int) (int, error) {
 	}
 	return count, nil
 }
+
+// IsVehicle returns if the item is a vehicle or not
+func (i *Item) IsVehicle() bool {
+	return i.Category.Name == "Cars and Heavy Vehicles"
```
briskt, line 1201: Our philosophy is to never assume that child items are hydrated. So I would recommend calling `Load` on the item first.

jason-jackson: Given I am adding premium factor as a field in the item_categories table
Then I will just remove this function until we actually need it.

application/migrations/20230906184608_add_item_category_billing_period_premium.up.fizz
```diff
@@ -0,0 +1,2 @@
+add_column("item_categories", "billing_period", "int", {"null": true})
```
jason-jackson, line 1: Would this be a better alternative?

```suggestion
add_column("item_categories", "billing_period", "int", {"default": 12})
```

briskt: Probably, yes. I prefer to avoid nullable columns when possible. I'm forgetting, though, does that work if there is already data in the table?

application/models/itemcategory.go
```diff
@@ -34,6 +34,8 @@ type ItemCategory struct {
 	Status           api.ItemCategoryStatus `db:"status" validate:"itemCategoryStatus"`
 	AutoApproveMax   int                    `db:"auto_approve_max" validate:"min=0"`
 	RequireMakeModel bool                   `db:"require_make_model"`
+	PremiumFactor    *float64               `db:"premium_factor"`
```
briskt, line 37:
```suggestion
	PremiumFactor    nulls.Float64         `db:"premium_factor"`
```

application/migrations/20230906184608_add_item_category_billing_period_premium.up.fizz
jason-jackson: Yes, default is how to do add a non-null field. See https://stackoverflow.com/questions/512451/how-can-i-add-a-column-that-doesnt-allow-nulls-in-a-postgresql-database.

briskt (approved):
Looks good

PR #533, opened 2023-09-06
fix swagger docs to play nicely with gofmt

briskt:
### Added
- Add swagger.json to version control to monitor for changes. (You can look at the commits after the first one (5aea97c) to identify swagger changes made in this PR.)

### Fixed
- Formatted swagger yaml specs as code blocks to prevent gofmt from changing them. (Start each line with `//` and a single tab character.)
- Removed empty lines that would trigger gofmt to add a header format character (`#`).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,7 +33,6 @@ application/emailtester/*.go
 .vscode/
 application/grifts/private.go
 application/migrations/schema.sql
-application/swagger.json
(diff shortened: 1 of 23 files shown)


jason-jackson (approved):
Thanks for doing this. Looks good, though I mostly just quickly glanced over the swagger docs.

PR #535, opened 2023-09-06
define init sequence in preparation for Go 1.21

briskt:
### Changed
- Perform all package initialization from `main`. This avoids any possible bugs introduced by the new initialization sequence in Go 1.21.

### DIFF
--- a/application/domain/domain.go
+++ b/application/domain/domain.go
@@ -245,8 +245,8 @@ func readEnv() *EnvStruct {
 	env.PolicyMaxCoverage *= CurrencyFactor
 	env.DependentAutoApproveMax *= CurrencyFactor
 	env.PremiumMinimum *= CurrencyFactor
-	env.RepairThresholdString = fmt.Sprintf("%.2g%%", env.RepairThreshold*100)
-	env.DeductibleMinimumString = fmt.Sprintf("%.2g%%", env.Deductible*100)
+	env.RepairThresholdString = fmt.Sprintf("%.2g%%", env.RepairThreshold*CurrencyFactor)
+	env.DeductibleMinimumString = fmt.Sprintf("%.2g%%", env.Deductible*CurrencyFactor)
 
 	//  Set an arbitrary but reasonable minimum lifetime for policy strikes
 	if env.StrikeLifetimeMonths < 2 {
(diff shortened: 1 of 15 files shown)


application/domain/domain.go
```diff
@@ -233,25 +232,28 @@ func init() {
...
-	Env.PremiumMinimum *= CurrencyFactor
-	Env.RepairThresholdString = fmt.Sprintf("%.2g%%", Env.RepairThreshold*100)
-	Env.DeductibleMinimumString = fmt.Sprintf("%.2g%%", Env.Deductible*100)
+	checkSamlConfig(env)
+
+	env.PolicyMaxCoverage *= CurrencyFactor
+	env.DependentAutoApproveMax *= CurrencyFactor
+	env.PremiumMinimum *= CurrencyFactor
+	env.RepairThresholdString = fmt.Sprintf("%.2g%%", env.RepairThreshold*100)
+	env.DeductibleMinimumString = fmt.Sprintf("%.2g%%", env.Deductible*100)
```
jason-jackson, line 249:
```suggestion
	env.RepairThresholdString = fmt.Sprintf("%.2g%%", env.RepairThreshold*CurrencyFactor)
	env.DeductibleMinimumString = fmt.Sprintf("%.2g%%", env.Deductible*CurrencyFactor)
```

application/domain/domain.go
```diff
@@ -130,8 +129,10 @@ const EnvStaging = "staging"
 const EnvTest = "test"
 
 // Env Holds the values of environment variables
-var Env struct {
-	GoEnv                      string `ignored:"true"`
+var Env = readEnv()
+
+type EnvStruct struct {
+	GoEnv                      string `default:"development" split_words:"true"`
```
jason-jackson, line 135: Do you know if it's `GOENV` or `GO_ENV`.  I was thinking it was the former, but we were using the latter?

jason-jackson (approved):

application/domain/domain.go
jason-jackson: My thoughts were related to `GOOS`/`GOARCH`.  `GO_ENV` is a Buffalo thing.

briskt: It's `GO_ENV`. It's not a Go thing (like `GOOS`), from what I can recall. It's only defined in Buffalo.

PR #537, opened 2023-09-07
rollback changes made to experiment with env var config

briskt:


jason-jackson (approved):
Remove ENV_ENCRYPTED_VAR from encrypted file

PR #539, opened 2023-09-07
fix TestNetSuite_Export test by ignoring the line number

briskt:
### Fixed
- Fix the TestNetSuite_Export test by ignoring the line number.

Issue: [CVR-671](https://itse.youtrack.cloud/issue/CVR-671)

### DIFF
--- a/application/fin/netsuite_test.go
+++ b/application/fin/netsuite_test.go
@@ -5,7 +5,6 @@ import (
 	"bytes"
 	"fmt"
 	"io"
-	"regexp"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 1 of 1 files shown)


application/fin/netsuite_test.go
```diff
@@ -100,16 +96,12 @@ func TestNetSuite_Export(t *testing.T) {
...
+			want = transaction2Row[12:]
 		}
 
-		// Replace the first instance of "<number>," with "\d,"
-		// So that map order will not affect test
-		// For example, 2023080000001 becomes 202308000000\d
-		want = re.ReplaceAllString(want, `$1\d,$2`)
-
-		require.Regexp(t, netSuiteHeader+want, string(body))
+		require.Contains(t, string(body), want)
```
jason-jackson, line 105: This seems a bit cleaner, and shouldn't break if we decide to change csv contents later (assuming we keep id as the first field).

```suggestion
		// don't try to compare the row number since we can't guarantee the transaction batch ordering
		row := transaction1Row
		if name == "bar" {
			row = transaction2Row
		}
		_, want, _ := strings.Cut(row, ",")

		require.Contains(t, string(body), want)
```

jason-jackson (approved):

PR #544, opened 2023-09-13
changed monthly premium calculation

briskt:
### Changed
- Changed `CalculateMonthlyPremium` to assume `PremiumFactor` is an annual percentage.
- Return the premium amount from `CalculateMonthlyPremium` for any item, not just those with a category that has a `PremiumFactor` value.

### DIFF
--- a/application/models/item.go
+++ b/application/models/item.go
@@ -704,7 +704,6 @@ func (i *Item) LoadPolicyMembers(tx *pop.Connection, reload bool) {
 	i.Policy.LoadMembers(tx, reload)
 }
 
-// TODO: split this into two functions (and add a `reload` parameter)
 // Load - a simple wrapper method for loading child objects
 func (i *Item) Load(tx *pop.Connection) {
 	if i.Category.ID == uuid.Nil {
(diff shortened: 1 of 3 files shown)


application/models/item.go
```diff
@@ -843,16 +843,17 @@ func (i *Item) CalculateProratedPremium(t time.Time) api.Currency {
...
-func (i *Item) CalculateMonthlyPremium(tx *pop.Connection) *api.Currency {
+// CalculateMonthlyPremium returns the rounded product of the item's CoverageAmount and the category's
+// PremiumFactor divided by 12
+func (i *Item) CalculateMonthlyPremium(tx *pop.Connection) api.Currency {
 	i.Load(tx)
-
 	if i.Category.PremiumFactor.Valid {
-		premium := api.Currency(math.Round(float64(i.CoverageAmount) * i.Category.PremiumFactor.Float64))
-		return &premium
+		premium := api.Currency(math.Round(float64(i.CoverageAmount) * i.Category.PremiumFactor.Float64 / 12))
```
jason-jackson, line 851: Use `i.BillingPeriod` instead of 12 (might need to uncomment it)

jason-jackson (approved):

application/models/item.go
briskt: Actually, I don't think so, because the name of the function is `CalculateMonthlyPremium`. The caller needs to choose the correct function based on the value of `BillingPeriod`.

PR #545, opened 2023-09-13
migrate premium factor into the item_categories table

briskt:
### Changed
- Copy the configured premium factor into all rows of the item_categories table that have NULL in the `premium_factor` column.
- Use a rounded premium factor in the `ItemCategory` API object.


application/models/models.go
```diff
@@ -173,6 +173,17 @@ func Init() {
 	if _, err = getRandomToken(); err != nil {
 		panic(fmt.Errorf("error using crypto/rand ... %v", err))
 	}
+
+	itemCategories := ItemCategories{}
+	if err := DB.Where("premium_factor IS NULL").All(&itemCategories); err != nil {
```
jason-jackson, line 178: Use `domain.IsOtherThanNoRows`, as not finding any NULLs should not break application

jason-jackson (changes requested):

application/models/item.go
```diff
@@ -303,10 +305,10 @@ func (i *Item) SafeDeleteOrInactivate(ctx context.Context) error {
 			return nil
 		}
 
-		return i.CreateLedgerEntry(Tx(ctx), LedgerEntryTypeCoverageRefund, i.calculateCancellationCredit(now))
+		return i.CreateLedgerEntry(tx, LedgerEntryTypeCoverageRefund, i.calculateCancellationCredit(tx, now))
 
 	case api.ItemCoverageStatusDraft, api.ItemCoverageStatusRevision, api.ItemCoverageStatusPending:
-		tx := Tx(ctx)
+		tx := tx
```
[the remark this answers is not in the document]
briskt: Haha, Jetbrains failed me on this one.

application/models/models.go
```diff
@@ -173,6 +173,17 @@ func Init() {
...
+		panic(fmt.Sprintf("failed to query item_categories: %s", err))
+	}
+	for i := range itemCategories {
+		itemCategories[i].PremiumFactor = nulls.NewFloat64(domain.Env.PremiumFactor)
+		if err := DB.UpdateColumns(&itemCategories[i], "premium_factor", "updated_at"); err != nil {
+			panic(fmt.Sprintf("failed to set item_categories.premium_factor: %s", err))
+		}
+	}
```
jason-jackson, line 186: Noting that this makes `domain.Env.PremiumFactor` obsolete, which is not necessarily a bad thing, but it would be good to have a way to update the item's premium factors somehow (UI issue, do we have a ticket for that?)

application/models/item.go
```diff
@@ -832,14 +836,21 @@ func (i *Items) ConvertToAPI(tx *pop.Connection) api.Items {
...
+// PremiumFactor
+func (i *Item) CalculateAnnualPremium(tx *pop.Connection) api.Currency {
+	i.Load(tx)
+	if i.Category.PremiumFactor.Valid {
+		premium := api.Currency(math.Round(float64(i.CoverageAmount) * i.Category.PremiumFactor.Float64))
+		return premium
+	} else {
+		premium := api.Currency(math.Round(float64(i.CoverageAmount) * domain.Env.PremiumFactor))
+		return premium
+	}
```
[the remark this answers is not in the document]
jason-jackson: Also noting that because we add `PremiumFactor` in Init, `i.Category.PremiumFactor.Valid` should always be `true`

application/models/models.go
briskt: Good catch

application/models/models.go
briskt: I'm content to leave it as an Adminer task for the forseeable future.

application/models/item.go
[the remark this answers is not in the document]
briskt: Should be, yes. But we can't convert it to a regular float64 until later, and it can't hurt to sanity check until then.

jason-jackson (approved):
If a UI ticket for adding premium factor doesn't exist, it should be created.

PR #549, opened 2023-09-14
calculate minimum coverage based on variable premium factor

briskt:


application/models/item.go
```diff
@@ -554,6 +553,18 @@ func (i *Item) SubmitForApproval(ctx context.Context) error {
...
 
+func (i *Item) calculateMinimumCoverage(tx *pop.Connection) int {
+	i.LoadCategory(tx, false)
+	billingPeriod := i.Category.BillingPeriod
+	if billingPeriod < 1 {
+		billingPeriod = 1
+	}
+	if billingPeriod > 12 {
+		billingPeriod = 12
+	}
```
jason-jackson, line 564: I'd probably prefer this logic in Category model, such as in `i.Category.getBillingPeriod()` (or make it public if needed)

jason-jackson (approved):

PR #550, opened 2023-09-14
update mail templates to account for monthly premiums

briskt:


application/templates/mail/_item_card.plush.html
```diff
@@ -44,14 +44,16 @@ <h4 style="margin:0; padding:0;">Unique ID</h4>
 	) %>
 
 	<table style="padding: 0 0 1em 0;">
-		 <tr>
-			<th style="text-align: left; padding: 1em 1em 0 0;">Covered value</th>
+		<tr>
+			<th style="text-align: left; padding: 1em 1em 0 0;">Value</th>
+		</tr>
+		<tr>
```
jason-jackson, line 50: This doesn't look right, I think both `th`  should be in same `tr`.

```suggestion
```

jason-jackson (approved):

application/templates/mail/_item_card.plush.html
briskt: Alex moved it around a bit in CVR-684

briskt: I saw the problem eventually, but not before merging. Fixed with a commit on develop.

PR #554, opened 2023-09-18
expand renewal process to include monthly billing

briskt:


application/models/policy.go
```diff
@@ -599,37 +599,49 @@ func (p *Policy) createInvite(tx *pop.Connection, invite api.PolicyUserInviteCre
...
+func (p *Policy) ProcessRenewals(tx *pop.Connection, date time.Time, billingPeriod int) error {
 	var items Items
 	if err := tx.Where("coverage_status = ?", api.ItemCoverageStatusApproved).
-		Where("paid_through_date < ?", domain.EndOfYear(year)).
+		Where("paid_through_date < ?", date).
 		Where("policy_id  = ?", p.ID).
+		Join("item_categories ic", "items.category_id = ic.id").
+		Where("ic.billing_period = ?", billingPeriod).
 		All(&items); err != nil {
 		return appErrorFromDB(err, api.ErrorQueryFailure)
```
jason-jackson, line 623: Noting that this will result in a Failure if no results, do we intend for that or should we deal with that separately?

jason-jackson (approved):

application/models/policy.go
briskt: I think it's appropriate. This shouldn't be called if there aren't any items to be billed since there is a separate endpoint to check the status. I suppose there may be a better way to do this, but it seems to work.

PR #555, opened 2023-09-18
two-tier premium

briskt:
### Added
- Added a second premium tier to the `ItemCategory` table/model.

### Changed
- Make the `PremiumFactor` field on `ItemCategory` not nullable.

### Removed
- Removed `domain.Env.PremiumFactor`.


application/migrations/20230918150113_add_premium_levels.up.fizz
```diff
@@ -0,0 +1,5 @@
+add_column("item_categories", "premium_factor_high", "real", {"default": 0})
+add_column("item_categories", "premium_threshold", "int", {"null": true})
+sql(`
+  ALTER TABLE "item_categories" ALTER "premium_factor" SET NOT NULL;
```
jason-jackson, line 4: Not sure if we can do this without either an intermediate production release or updating null values to a specific default first before this here.  For example:

```suggestion
  UPDATE "item_categories" SET "premium_factor" = 0.02 WHERE "premium_factor" IS NULL;
  ALTER TABLE "item_categories" ALTER "premium_factor" SET NOT NULL;
```

application/models/item.go
```diff
@@ -559,7 +559,8 @@ func (i *Item) SubmitForApproval(ctx context.Context) error {
 func (i *Item) calculateMinimumCoverage(tx *pop.Connection) int {
 	i.LoadCategory(tx, false)
 	billingPeriod := i.Category.getBillingPeriod()
-	return int(math.Round(0.5 / i.Category.PremiumFactor.Float64 / 12 * float64(billingPeriod)))
+	premiumFactor := i.Category.getPremiumFactor(i.CoverageAmount)
+	return int(math.Round(0.5 / premiumFactor / 12 * float64(billingPeriod)))
```
jason-jackson, line 563: This 12 most likely could be changed to BillingPeriodAnnual, but not 100%. It is how many months are in a year.

application/models/item.go
```diff
@@ -889,13 +886,8 @@ func (i *Item) CalculateProratedPremium(tx *pop.Connection, t time.Time) api.Cur
...
 	i.LoadCategory(tx, false)
-	if i.Category.PremiumFactor.Valid {
-		premium := api.Currency(math.Round(float64(i.CoverageAmount) * i.Category.PremiumFactor.Float64 / 12))
-		return premium
-	} else {
-		premium := api.Currency(math.Round(float64(i.CoverageAmount) * domain.Env.PremiumFactor / 12))
-		return premium
-	}
+	premiumFactor := i.Category.getPremiumFactor(i.CoverageAmount)
+	return api.Currency(math.Round(float64(i.CoverageAmount) * premiumFactor / 12))
```
jason-jackson, line 890: This 12 most likely could be changed to BillingPeriodAnnual, but not 100%. It is how many months are in a year.

application/migrations/20230918150113_add_premium_levels.up.fizz
jason-jackson: It not being released is what makes me think it will be a problem
1. `20230906184608_add_item_category_billing_period_premium.up.fizz` gets run and creates `premium_factor` with NULL
2. `20230918150113_add_premium_levels.up.fizz` gets run and tries to set it to `NOT NULL`, but all the values in production will be `NULL`.

It should be fine in staging, but not in production as there was an intermittent PR that updated the values to the env variable (this code and env variable is being removed in this PR, not that keeping it would help in this case either as the migrations would happen before Init, I think.

briskt: Oh, right. Thanks.

jason-jackson (approved):
Do we need this? I was under the impression that high/low premium values were not going to be used for vehicles.

briskt:
This is not the final decision on vehicle premium calculation.

PR #556, opened 2023-09-19
correctly handle cancellation on items billed monthly

briskt:


application/models/item.go
```diff
@@ -356,15 +360,17 @@ func (i *Item) isNewEnough() bool {
...
-	// If now it's January and the item was created before this year
-	//   set its CoverageEndDate to the current day.
-	// Otherwise, set it to the end of the current month.
-	if i.shouldGiveFullYearRefund(t) {
+	tx := Tx(ctx)
+	i.LoadCategory(tx, false)
+
+	// If the item was created before this year, and it qualifies for a full-year refund, set
+	// its CoverageEndDate to the current day. Otherwise, set it to the end of the current month.
+	if i.Category.BillingPeriod == domain.BillingPeriodAnnual && i.shouldGiveFullYearRefund(t) {
```
jason-jackson, line 373: I wonder if `i.Category.BillingPeriod == domain.BillingPeriodAnnual` should be moved into `i.shouldGiveFullYearRefund(t)`

jason-jackson (approved):

application/models/item.go
briskt: Perhaps, but that would mean passing tx into it so it can call `LoadCategory`. Do you think it's worth it?

jason-jackson: Probably not at this point.

PR #557, opened 2023-09-20
use correct calculation for premium on new coverage

briskt:


application/models/item.go
```diff
@@ -1037,26 +1059,27 @@ func (i *Item) CreateLedgerEntry(tx *pop.Connection, entryType LedgerEntryType,
...
+	date = date.Truncate(24 * time.Hour)
+	if date == i.PaidThroughDate {
+		return nil
 	}
 
+	i.PaidThroughDate = date
+	if err := tx.UpdateColumns(i, "paid_through_date", "updated_at"); err != nil {
+		return appErrorFromDB(err, api.ErrorUpdateFailure)
+	}
 	return nil
```
jason-jackson, line 1083:
```suggestion
	err := tx.UpdateColumns(i, "paid_through_date", "updated_at")
	return appErrorFromDB(err, api.ErrorUpdateFailure)
```

jason-jackson (approved):

application/models/item.go
briskt: I realize this works, and has been done quite a bit, but I'm less comfortable with it now. Calling `appErrorFromDB` regardless of `err != nil` is not as idiomatic.

PR #558, opened 2023-09-21
add a new risk category for Vehicles

briskt:
### Added
- Add a new risk category for Vehicles

### Removed
- Removed `policy_max` column from the `risk_categories` table. It was never used.

### DIFF
--- a/application/actions/app.go
+++ b/application/actions/app.go
@@ -204,10 +204,11 @@ func App() *buffalo.App {
 
 		// config
 		configGroup := app.Group("/config")
-		configGroup.Middleware.Skip(AuthZ, claimIncidentTypes, itemCategoriesList, entityCodesList, countries)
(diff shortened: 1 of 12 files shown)


application/migrations/20230921155455_add_vehicle_risk_category.up.fizz
```diff
@@ -0,0 +1,6 @@
+drop_column("risk_categories", "policy_max")
```
hobbitronics, line 1: is policy_max still meant to be here?

application/actions/app.go
```diff
@@ -216,6 +216,7 @@ func App() *buffalo.App {
 
 		// entity codes
 		entityCodesGroup := app.Group(entityCodesPath)
+		configGroup.Middleware.Skip(AuthZ, entityCodesList)
```
jason-jackson, line 219:
```suggestion
		entityCodesGroup.Middleware.Skip(AuthZ, entityCodesList)
```

jason-jackson (approved):

PR #563, opened 2023-09-22
add a new column to item_categories for use as a key field in the UI

briskt:


application/migrations/20230922215842_add_item_catogory_key.up.fizz
```diff
@@ -0,0 +1 @@
+add_column("item_categories", "key", "string", {"default": ""})
```
jason-jackson, line 1: Probably would be good to add unique index too.

jason-jackson (approved):

PR #566, opened 2023-09-25
expose Item Year to the API

briskt:


application/api/items.go
```diff
@@ -167,6 +167,9 @@ type ItemCreate struct {
 	// serial number
 	SerialNumber string `json:"serial_number"`
 
+	// year, numeric, optional. Omitting this field will erase the stored value.
```
forevermatt, line 170: Looks like this has a copy/paste error. Since this is an ItemCreate call, I assume there's no stored value to be erased.

application/swagger.json
```diff
@@ -2779,6 +2785,12 @@
           "description": "serial number",
           "type": "string",
           "x-go-name": "SerialNumber"
+        },
+        "year": {
+          "description": "year, numeric, optional. Omitting this field will erase the stored value.",
```
forevermatt, line 2790: My guess is this is generated from the other file, but this has the same copy/paste typo (i.e. this is a new item, so there is no stored value to erase).

briskt: That's correct.

application/api/items.go
briskt: That's a good observation. I'll remove that sentence.

application/actions/items_test.go
```diff
@@ -914,6 +915,7 @@ func (as *ActionSuite) Test_NewItemFromApiInput() {
 		Make:                "Minolta",
 		Model:               "Max",
 		SerialNumber:        "MM1234",
+		Year:                models.NullsIntToPointer(nulls.NewInt(1999)),
```
jason-jackson, line 918: Seems a bit roundabout, but not really seeing a different one-line alternative to this. Would it be cleaner to go ahead and declare `year := 1999` and set this to `Year: &year`?

jason-jackson (approved):
Looks good, main "issue" I have is with the roundabout way to make the year pointer, but that's still minor.

application/models/models.go
```diff
@@ -441,3 +441,18 @@ func GetHHID(staffID string) string {
...
 	return v.ID
 }
+
+func NullsIntToPointer(i nulls.Int) *int {
+	if i.Valid {
+		value := i.Int
+		return &value
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
briskt: Both of those return a pointer to the original value. I didn't want to expose that to the downstream code.

application/models/models.go
```diff
@@ -441,3 +441,18 @@ func GetHHID(staffID string) string {
...
+
+func PointerToNullsInt(i *int) nulls.Int {
+	if i == nil {
+		return nulls.Int{}
+	}
+	return nulls.NewInt(*i)
```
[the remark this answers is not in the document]
briskt: Yeah, I think there are things that could be added to the `gobuffalo/nulls` package. I wonder if they'd accept a PR to add them.

application/actions/items_test.go
briskt: Hmm, maybe. I kind of did this without thinking very much. I may just leave it to keep forward progress.

PR #568, opened 2023-09-26
import vehicle data for household coverage

briskt:


application/models/policy.go
```diff
@@ -670,3 +679,150 @@ func (p *Policy) CreateRenewalLedgerEntry(tx *pop.Connection, riskCategoryID uui
...
+		}
+		response.PoliciesCreated += policiesCreated
+		response.ItemsCreated += itemsCreated
+	}
+
+	response.LinesProcessed = n
+	return response, nil
+}
+
+func importPolicy(tx *pop.Connection, csvLine []string, catID uuid.UUID, now time.Time) (int, int, error) {
```
jason-jackson, line 722:
```suggestion
func importVehiclePolicy(tx *pop.Connection, csvLine []string, catID uuid.UUID, now time.Time) (int, int, error) {
```

application/models/policy.go
```diff
@@ -670,3 +679,150 @@ func (p *Policy) CreateRenewalLedgerEntry(tx *pop.Connection, riskCategoryID uui
 
 	return nil
 }
+
+func ImportPolicies(tx *pop.Connection, file io.Reader) (api.PoliciesImportResponse, error) {
```
jason-jackson, line 683:
```suggestion
func ImportVehiclePolicies(tx *pop.Connection, file io.Reader) (api.PoliciesImportResponse, error) {
```

application/models/policy.go
```diff
@@ -670,3 +679,150 @@ func (p *Policy) CreateRenewalLedgerEntry(tx *pop.Connection, riskCategoryID uui
...
+	if err != nil {
+		return 0, fmt.Errorf("invalid covered value %q: %w", s, err)
+	}
+	if value < 1 || value > 100_000 {
+		return 0, fmt.Errorf("invalid covered value %d", value)
+	}
+	return value * domain.CurrencyFactor, nil
+}
+
+func parseVehicleYear(s string) (int, error) {
```
jason-jackson, line 813: Would it be beneficial to create a check to make sure we don't accidentally call this without updating it later?

Such as:
```
	if time.Now().UTC().Year() >= 2050 {
		return 0, fmt.Errorf("parseVehicleYear has hardcoded dates, please verify logic")
	}
```

Or could make it dynamic, but I'm not sure that's worth the thought process to implement.

jason-jackson (approved):
Looks good, not sure about names, but I think it could be kept as is if desired.

application/models/policy.go
briskt: I'm not especially concerned about making sure this is still operating without issue for over 25 years. I feel like there would be much bigger problems in that situation.

application/models/policy.go
briskt: Yes, that would be more accurate. This could go one of two directions: we delete this code after vehicle coverage is launched, or we keep it with the idea that it might be useful for adding time machine coverage at a later date. I think the former is probably more likely.

PR #570, opened 2023-09-27
implement minimum premium

briskt:


application/models/itemcategory.go
```diff
@@ -35,6 +35,8 @@ type ItemCategory struct {
 	HelpText         string                 `db:"help_text"`
 	Status           api.ItemCategoryStatus `db:"status" validate:"itemCategoryStatus"`
 	AutoApproveMax   int                    `db:"auto_approve_max" validate:"min=0"`
+	MinimumPremium   int                    `db:"minimum_premium" validate:"min=0"`
```
jason-jackson, line 38: Could this be `api.Currency` instead of `int`? Not sure if that would break anything. I'm fine with int too, just curious.

jason-jackson (approved):
Just a minor question, not sure if it is worth investigating.

application/models/itemcategory.go
briskt: I think it could be. We haven't used `api.Currency` everywhere we could. Maybe a task for later.

PR #571, opened 2023-09-28
don't use CurrencyFactor on percentages

briskt:


jason-jackson (approved):
Sorry I missed that these were percentages before.

application/domain/domain.go
```diff
@@ -255,8 +255,8 @@ func readEnv() *EnvStruct {
 	env.PolicyMaxCoverage *= CurrencyFactor
 	env.DependentAutoApproveMax *= CurrencyFactor
 	env.PremiumMinimum *= CurrencyFactor
-	env.RepairThresholdString = fmt.Sprintf("%.2g%%", env.RepairThreshold*CurrencyFactor)
-	env.DeductibleMinimumString = fmt.Sprintf("%.2g%%", env.Deductible*CurrencyFactor)
+	env.RepairThresholdString = fmt.Sprintf("%.2g%%", env.RepairThreshold*100)
+	env.DeductibleString = fmt.Sprintf("%.2g%%", env.Deductible*100)
```
jason-jackson, line 259:
```suggestion
	env.DeductibleMinimumString = fmt.Sprintf("%.2g%%", env.Deductible*100)
```

briskt: Oh, thanks. I had made this change after renaming that field, but forgot to set it back.

PR #573, opened 2023-09-29
move the item_categories.premium_factor patch out of init()

briskt:
since init code runs before migrations, this method broke any time a migration changed the item_categories table

### DIFF
--- a/application/models/models.go
+++ b/application/models/models.go
@@ -168,17 +168,6 @@ func init() {
 		panic(fmt.Errorf("error using crypto/rand ... %v", err))
 	}
 
-	//itemCategories := ItemCategories{}
-	//if err := DB.Where("premium_factor IS NULL").All(&itemCategories); domain.IsOtherThanNoRows(err) {
-	//	panic(fmt.Sprintf("failed to query item_categories: %s", err))
-	//}
-	//for i := range itemCategories {
-	//	itemCategories[i].PremiumFactor = nulls.NewFloat64(domain.Env.PremiumFactor)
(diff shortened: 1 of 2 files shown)


application/models/models.go
```diff
@@ -194,6 +183,19 @@ func init() {
...
+	itemCategories := ItemCategories{}
+	if err := DB.Where("premium_factor IS NULL").All(&itemCategories); domain.IsOtherThanNoRows(err) {
+		panic(fmt.Sprintf("failed to query item_categories: %s", err))
+	}
+	for i := range itemCategories {
+		itemCategories[i].PremiumFactor = nulls.NewFloat64(domain.Env.PremiumFactor)
+		if err := DB.UpdateColumns(&itemCategories[i], "premium_factor", "updated_at"); err != nil {
+			panic(fmt.Sprintf("failed to set item_categories.premium_factor: %s", err))
+		}
+	}
```
jason-jackson, line 196: I wonder which is better, this or the following (move `UpdateColumns` out of the loop):

```suggestion
	for i := range itemCategories {
		itemCategories[i].PremiumFactor = nulls.NewFloat64(domain.Env.PremiumFactor)
	}
	if err := DB.UpdateColumns(itemCategories, "premium_factor", "updated_at"); err != nil {
		panic(fmt.Sprintf("failed to set item_categories.premium_factor: %s", err))
	}
```

jason-jackson (approved):

application/models/models.go
briskt: That's a good optimization, though I don't know how much it matters. Either we do the loop or Pop does. I'll try to remember this next time, if only because the code is a little bit cleaner because of the decreased indentation.

PR #575, opened 2023-10-02
set the item's coverage_start_date correctly

briskt:
### Fixed
- Update an item's `CoverageStartDate` when coverage is approved.
- Delay coverage to the next month on an item billed monthly if the monthly cutoff date has passed.

### Removed
- Remove `CoverageStartDate` and `CoverageEndPoint` from the `ItemCreate` struct, which is used on the PolicyItemsCreate endpoint because these dates were not properly supported at the time of creating a new Item.

---

### Internal changes:

- `Item` `Approve` takes the current time for better test coverage and uniform handling of the "current time".
- `Item` `Approve` does not accept a boolean to control whether the Item Approved event should be emitted. This boolean was always true except in test code.

### DIFF
--- a/application/actions/items_test.go
+++ b/application/actions/items_test.go
@@ -148,12 +148,11 @@ func (as *ActionSuite) Test_ItemsCreate() {
 		SerialNumber:        "MM1234",
 		CoverageAmount:      101,
 		CoverageStatus:      api.ItemCoverageStatusDraft,
-		CoverageStartDate:   "2006-01-03",
 		AccountablePersonID: policyCreator.ID,
 	}
 
-	badItemDate := goodItem
-	badItemDate.CoverageStartDate = "1/1/2020"
+	missingPerson := goodItem
+	missingPerson.AccountablePersonID = uuid.Nil
 
 	tests := []struct {
 		name       string
@@ -184,9 +183,9 @@ func (as *ActionSuite) Test_ItemsCreate() {
 			name:       "bad request",
 			actor:      policyCreator,
 			policy:     policy,
-			newItem:    badItemDate,
+			newItem:    missingPerson,
 			wantStatus: http.StatusBadRequest,
-			wantInBody: []string{api.ErrorItemInvalidCoverageStartDate.String()},
+			wantInBody: []string{api.ErrorItemNullAccountablePerson.String()},
 		},
 		{
 			name:       "ok",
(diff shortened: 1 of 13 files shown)


application/swagger.json
```diff
@@ -1260,6 +1260,35 @@
         }
       }
     },
+    "/policies/{id}/imports": {
```
briskt, line 1263: What do you think, should I remove swagger.json from the repo again? It's hard to remember to update it when it actually needs to change.

jason-jackson: Maybe try to have it be rebuilt every time dev is built? Seems like that is a bit overkill. How much added value do you think it brings?

application/actions/items_test.go
```diff
@@ -291,7 +290,7 @@ func (as *ActionSuite) Test_ItemsSubmit() {
 				`"serial_number":"` + revisionItem.SerialNumber,
 				// keeps revisionItem coverage_amount
 				fmt.Sprintf(`"coverage_amount":%v`, revisionItem.CoverageAmount),
-				`"coverage_start_date":"` + revisionItem.CoverageStartDate.Format(domain.DateFormat) + `"`,
+				`"coverage_start_date":"` + time.Now().Format(domain.DateFormat) + `"`,
```
jason-jackson, line 293:
```suggestion
				`"coverage_start_date":"` + time.Now().UTC().Format(domain.DateFormat) + `"`,
```

jason-jackson (approved):
I'm curious about events, but emitted events didn't really change, so not going to hold it up.

application/models/item.go
```diff
@@ -665,14 +667,12 @@ func (i *Item) AutoApprove(ctx context.Context) error {
...
 
 	i.StatusChange = ItemStatusChangeAutoApproved
-	return i.Approve(ctx, true)
+	return i.Approve(ctx, time.Now().UTC())
 }
 
 // Approve takes the item from Pending coverage status to Approved.
 // It assumes that the item's current status has already been validated.
-// Only emits an event for an email notification if requested.
-// (No need to emit it following an auto-approval which has already emitted an event.)
```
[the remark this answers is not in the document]
briskt: I don't understand the comment, actually. Both events need to fire. One is for sending an email to the customer and one is for sending an email to the steward.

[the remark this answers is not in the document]
jason-jackson: Ah, ok, carry on then.

PR #584, opened 2023-10-12
use Max helper function

briskt:
### Added
- Refactor: create and use a helper function `Max`

---

### Code Checklist
- [x] Run `gofmt`
- [x] Run `make swaggerspec`

### DIFF
--- a/application/domain/domain.go
+++ b/application/domain/domain.go
@@ -248,7 +248,7 @@ func readEnv() *EnvStruct {
 	env.DeductibleRateString = PercentString(env.DeductibleRate)
 
 	//  Set an arbitrary but reasonable minimum lifetime for policy strikes
-	env.StrikeLifetimeMonths = Min(2, env.StrikeLifetimeMonths)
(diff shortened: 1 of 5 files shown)


application/domain/domain.go
```diff
@@ -515,3 +513,11 @@ func checkSamlConfig(env *EnvStruct) {
 		panic("required SAML variable SamlSpPrivateKey is undefined")
 	}
 }
+
+// TODO: replace this with the builtin function after upgrading to Go 1.21
+func Max(a, b int) int {
```
jason-jackson, line 518: I don't think this is worth it, but mentioning it anyways.

Alternately, could use a generic with `"golang.org/x/exp/constraints"` package or by defining own input (suggest something along the lines of `int|api.Currency`).

```suggestion
func Max[T constraints.Ordered](a, b T) T {
```

jason-jackson (approved):

application/domain/domain.go
briskt: It's a worthy mention, but I hope we won't be waiting long until Buffalo is ready for Go 1.21

PR #589, opened 2023-10-18
ensure the Claim TotalPayout is updated after adding a ClaimItem

briskt:
### Fixed
- Update the TotalPayout on a Claim when adding a ClaimItem. Previously, it was only updated when the Claim itself was updated.

---

### Code Checklist
- [x] Run `gofmt`
- [x] Run `make swaggerspec`

[CVR-709](https://itse.youtrack.cloud/issue/CVR-709)

### DIFF
--- a/application/models/testutils.go
+++ b/application/models/testutils.go
@@ -939,8 +939,8 @@ func Must(err error) {
 	}
 }
 
-// SameAppError verifies that the actual error contains an AppError and that the key and category match expected
-func SameAppError(t *testing.T, expected api.AppError, actual error) {
+// AssertSameAppError verifies that the actual error contains an AppError and that the key and category match expected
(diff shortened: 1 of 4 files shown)


application/models/testutils.go
```diff
@@ -937,3 +938,15 @@ func Must(err error) {
 		panic(err.Error())
 	}
 }
+
+// SameAppError verifies that the actual error contains an AppError and that the key and category match expected
+func SameAppError(t *testing.T, expected api.AppError, actual error) {
```
jason-jackson, line 943: Could make it a bit clearer that this is an assertion / requirement

```suggestion
func AssertSameAppError(t *testing.T, expected api.AppError, actual error) {
```

jason-jackson (approved):

PR #590, opened 2023-10-23
import team vehicles

briskt:
### Added
- Added support for importing team policies

---

### Code Checklist
- [x] Run `gofmt`
- [x] Run `make swaggerspec`

### DIFF
--- a/application/models/policy.go
+++ b/application/models/policy.go
@@ -722,14 +722,14 @@ func ImportPolicies(tx *pop.Connection, file io.Reader) (api.PoliciesImportRespo
 		return response, api.NewAppError(err, api.ErrorUnknown, api.CategoryInternal)
 	}
 
-	n := 2 // first line of data is row 2 in the spreadsheet
+	n := 0
 	for ; ; n++ {
 		csvLine, err := r.Read()
 		if err == io.EOF {
 			break
 		}
 		if err != nil {
-			err := fmt.Errorf("failed to read from policy CSV file on line %v: %w", n, err)
+			err := fmt.Errorf("failed to read from policy CSV file on row %d: %w", n+2, err)
 			return response, api.NewAppError(err, api.ErrorUnknown, api.CategoryUser)
 		}
 
@@ -739,14 +739,14 @@ func ImportPolicies(tx *pop.Connection, file io.Reader) (api.PoliciesImportRespo
 		}
 		policiesCreated, itemsCreated, err := importPolicy(tx, data, vehicleCategory.ID, time.Now().UTC())
 		if err != nil {
-			err := fmt.Errorf("error importing policy on CSV file line %v: %w", n, err)
+			err := fmt.Errorf("error importing policy on row %d: %w", n+2, err)
 			return response, api.NewAppError(err, api.ErrorUnknown, api.CategoryUser)
 		}
 		response.PoliciesCreated += policiesCreated
 		response.ItemsCreated += itemsCreated
 	}
 
-	response.LinesProcessed = n - 2
+	response.LinesProcessed = n
 	return response, nil
 }
 
--- a/application/models/policy.go
+++ b/application/models/policy.go
@@ -426,8 +426,23 @@ func (p *Policy) AddDependent(tx *pop.Connection, input api.PolicyDependentInput
 
 	dependent.FixTeamRelationship(*p)
 
+	// If an identical record exists, use that one and don't create a duplicate
+	if err := dependent.FindByName(tx); err != nil {
+		if domain.IsOtherThanNoRows(err) {
+			return PolicyDependent{}, err
+		}
+	} else {
+		if dependent.Relationship == input.Relationship &&
+			dependent.Country == input.Country &&
+			dependent.ChildBirthYear == input.ChildBirthYear {
+			return dependent, nil
+		}
+		err = errors.New("cannot create a new PolicyDependent with same Name as existing record")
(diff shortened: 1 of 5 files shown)


application/models/policy.go
```diff
@@ -714,91 +745,143 @@ func ImportPolicies(tx *pop.Connection, file io.Reader) (api.PoliciesImportRespo
...
-	err := p.FindByHouseholdID(tx, csvLine[Account_Number])
-	if err != nil {
-		if domain.IsOtherThanNoRows(err) {
-			return 0, 0, appErrorFromDB(err, api.ErrorQueryFailure)
+
+	if data[HouseholdID] == "" {
+		err := p.FindByTeamDetails(tx, data[Entity], data[Account], data[CostCenter], data[AccountDetail])
+		if err != nil {
+			if domain.IsOtherThanNoRows(err) {
+				return 0, 0, err
```
jason-jackson, line 788:
```suggestion
				return 0, 0, appErrorFromDB(err, api.ErrorQueryFailure)
```

application/models/policy.go
```diff
@@ -714,91 +745,143 @@ func ImportPolicies(tx *pop.Connection, file io.Reader) (api.PoliciesImportRespo
...
+	}
+
+	if !dependent.ID.IsNil() {
+		i.PolicyDependentID = nulls.NewUUID(dependent.ID)
+	}
+
 	if err := i.Create(tx); err != nil {
 		return 0, 0, appErrorFromDB(err, api.ErrorCreateFailure)
 	}
 	return policiesCreated, 1, nil
```
jason-jackson, line 879: I think `policiesCreated` variable isn't strictly needed, will either be `0` (error) or `1` (here).

jason-jackson (changes requested):
Looks mostly good. Only `n` comment is holding me back from approving it.

application/models/policy.go
```diff
@@ -683,18 +709,19 @@ func ImportPolicies(tx *pop.Connection, file io.Reader) (api.PoliciesImportRespo
...
 
 	var vehicleCategory ItemCategory
-	if err := tx.Where("name ILIKE '%vehicles%'").First(&vehicleCategory); err != nil {
+	if err := tx.Where("risk_category_id = ?", riskCategoryVehicleID).First(&vehicleCategory); err != nil {
 		err := fmt.Errorf("failed to find an item category for vehicles: %w", err)
 		return response, api.NewAppError(err, api.ErrorUnknown, api.CategoryInternal)
 	}
 
-	n := 0
+	n := 2 // first line of data is row 2 in the spreadsheet
```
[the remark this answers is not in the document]
jason-jackson: In this case importing headers should fail on those lines, so not really a risk of bad data being entered.

application/models/policy.go
```diff
@@ -683,18 +709,19 @@ func ImportPolicies(tx *pop.Connection, file io.Reader) (api.PoliciesImportRespo
...
 	for ; ; n++ {
```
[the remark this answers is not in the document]
jason-jackson: Actually, no, would need to be set in 3rd column, I retract this comment.

application/models/policy.go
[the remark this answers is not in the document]
briskt: The header row is imported above this point. The variable `n` is used in three places. The first two are to print the line number in an error message. The third is to report the number of lines imported. As you observed, the third one uses `n - 2` to account for this `n := 2` assignment. I could have changed the lines that use `n` in an error message, but decided to go this route. Since it is apparently confusing this way, I will change it.

application/models/policy.go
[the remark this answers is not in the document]
briskt: Is this an improvement in readability, or just condensed?

application/models/policy.go
briskt: Why would this be necessary? The `err` comes from `FindByTeamDetails` which passes the error through `appErrorFromDB`.

application/models/policy.go
briskt: Good call

application/models/policy.go
briskt: Oops, I had commits I hadn't pushed yet.

application/models/policy.go
```diff
@@ -718,87 +750,136 @@ func ImportPolicies(tx *pop.Connection, file io.Reader) (api.PoliciesImportRespo
...
+	startDate, err := time.Parse("1/2/2006 15:04", data[StartDate])
 	if err != nil {
-		return 0, 0, fmt.Errorf("invalid date %v: %w", csvLine[Start_Date], err)
+		startDate = time.Date(now.Year(), now.Month(), 1, 0, 0, 0, 0, time.UTC)
 	}
 
-	year, err := parseVehicleYear(csvLine[Veh_Year])
-	if err != nil {
-		return 0, 0, err
+	itemName := data[ItemName]
```
jason-jackson, line 842: I halfway wonder if we want to have a `Default` func in domain, not really necessary, but I see it would get used twice in this PR.

```
func Default[T comparable](a, b T) T {
	var d T
	if a == d {
		return b
	}
	return a
}
```

jason-jackson (approved):
Thanks for the changes! I do think it's clearer with n+2 for line number instead.

application/models/policy.go
briskt: Interesting idea. It's kind of like a `min` or `max`. Wonder if anyone has proposed it for inclusion as a builtin function.

PR #577, opened 2023-10-03
coverage starts on the date of approval

briskt:
### Fixed
- Coverage starts at the date of approval, not the first day of premium billing.

[CVR-719](https://itse.youtrack.cloud/issue/CVR-729)

---

### Code Checklist
- [x] Documentation (README, goswagger, etc.)
- [x] Unit tests created or updated
- [x] Run `gofmt`
- [x] Run `make swaggerspec`

PR #592, opened 2023-10-24
Release 1.10.0 - vehicle coverage

briskt:
### Added
- Monthly billing option
- Category-based premium and deductible
- Vehicle risk category
- Added risk category name to transaction description

### Fixed
- Ensure the Claim TotalPayout is updated after adding a ClaimItem #589
- Coverage starts on the day of item approval #577


db-init.sh
jason-jackson, line 1: Odd file.

jason-jackson (dismissed):

db-init.sh
briskt: It's not actually empty. I don't know why Github shows it like that. This change was setting the execute flag.

application/models/testutils.go
```diff
@@ -106,15 +109,9 @@ func CreateFileFixtures(tx *pop.Connection, n int, createdByID uuid.UUID) Fixtur
...
-	}
-	if config.ItemsPerPolicy < 1 {
-		config.ItemsPerPolicy = 1
-	}
-	if config.ItemsPerPolicy < config.ClaimsPerPolicy*config.ClaimItemsPerClaim {
-		config.ItemsPerPolicy = config.ClaimsPerPolicy * config.ClaimItemsPerClaim
-	}
+	config.NumberOfPolicies = domain.Max(1, config.NumberOfPolicies)
+	config.ItemsPerPolicy = domain.Max(1, config.ItemsPerPolicy)
+	config.ItemsPerPolicy = domain.Max(config.ItemsPerPolicy, config.ClaimsPerPolicy*config.ClaimItemsPerClaim)
```
hobbitronics, line 114: is it preferable to have a spaces around the `*` or does that make the line break and harder to read?
```

**Review by hobbitronics — DISMISSED** [review]

File: application/actions/actions_test.go [hunk 2]
```diff
@@ -9,6 +9,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/silinternational/cover-api/domain"
```
hobbitronics, line 12: Why did the build fail for this change?
```Build Error: (step: deploy_and_scan_deploy_staging_web) container exited with a 1 code
ecsdeploy
ERROR: New task definition not running within 300 seconds```

File: application/models/testutils.go (hunk 1 above)
Comment (briskt) [reply application/models/testutils.go:114]: This is the formatting controlled by the gofmt tool, which is the standard Go formatter. If the multiplication was the only operator on the line it would put spaces around it. Like:
```go
	numberOfClaimItems := config.ClaimsPerPolicy * config.ClaimItemsPerClaim
	config.ItemsPerPolicy = domain.Max(config.ItemsPerPolicy, numberOfClaimItems)
```
I think this does look better, so I might make that change.

application/actions/actions_test.go
briskt: I am in the process of doing a database restore on the staging instance. It failed because the database was inaccessible at the time.

jason-jackson (approved):

application/actions/app.go
```diff
@@ -164,12 +168,15 @@ func App() *buffalo.App {
 		ledgerReportGroup := app.Group(ledgerReportPath)
 		// AuthZ is implemented in the handlers
 		ledgerReportGroup.Middleware.Skip(AuthZ, ledgerAnnualRenewalStatus, ledgerAnnualRenewalProcess)
+		ledgerReportGroup.Middleware.Skip(AuthZ, ledgerMonthlyRenewalStatus, ledgerMonthlyRenewalProcess)
```
forevermatt, line 171: I'm curious... why the skip-authorization for ledger renewal process and status calls?

application/actions/app.go
```diff
@@ -226,8 +234,10 @@ func App() *buffalo.App {
 
 		// policies
 		policiesGroup := app.Group(policiesPath)
+		policiesGroup.Middleware.Skip(AuthZ, policiesImport)
```
forevermatt, line 237: Similarly here, why skip authorization for the import-policies endpoint?

application/actions/app.go
briskt: The AuthZ middleware only supports API paths that look like `/resource/{id}|/resource/{id}|`. These and the `/policies/import` path don't fit that, so they have to be authorized within the handler.

PR #593, opened 2023-10-24
print currency using localization library

briskt:
### Changed
- Print currency using localization library, e.g. 1,000.00

---

### Code Checklist
- [x] Run `gofmt`
- [x] Run `make swaggerspec`

### DIFF
--- a/application/api/api_test.go
+++ b/application/api/api_test.go
@@ -151,6 +151,16 @@ func (ts *TestSuite) TestCurrency_String() {
 			c:    -100,
 			want: "-1.00",
 		},
+		{
+			name: "-100000",
+			c:    -100000,
+			want: "-1,000.00",
+		},
+		{
+			name: "100000",
+			c:    100000,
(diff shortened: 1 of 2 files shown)


application/api/api_test.go
```diff
@@ -151,6 +151,16 @@ func (ts *TestSuite) TestCurrency_String() {
 			c:    -100,
 			want: "-1.00",
 		},
+		{
+			name: "-100000",
+			c:    -100000,
+			want: "-1,000.00",
```
forevermatt, line 157: I don't know which is preferred, but on the UI side we're currently showing negative values as `($1,000.00)` rather than `-$1,000.00`. I'm also not sure where the currency symbol will show up in this change to the API-side code.

briskt: Good observation, the $ would appear before the negative symbol.

briskt:
This will affect more that just email messages. Closing due to unwanted complexity, and will revisit later.

PR #595, opened 2023-10-25
predate the renewal transaction to help with report generation

briskt:
### Changed
- Predate renewal transactions so the monthly report doesn't omit them if the report is run on the same day as the renewals.

---

### Code Checklist
- [x] Run `gofmt`
- [x] Run `make swaggerspec`

[CVR-762](https://itse.youtrack.cloud/issue/CVR-762)


jason-jackson (commented):
I could see this causing some confusion (I ran it on X date, but I'm seeing it on X-1 date), maybe UI should add a comment as well.

PR #599, opened 2023-10-26
fix policy search

briskt:
### Fixed
- Fixed the policy search feature. It was not including policies with no users.

---

### Code Checklist
- [x] Run `gofmt`
- [x] Run `make swaggerspec`

[CVR-777](https://itse.youtrack.cloud/issue/CVR-777)

### DIFF
--- a/application/models/policy.go
+++ b/application/models/policy.go
@@ -388,13 +388,13 @@ func scopeSearchPolicies(searchText string) pop.ScopeFunc {
 	return func(q *pop.Query) *pop.Query {
 		return q.Where("policies.id IN ("+
 			"SELECT policies.id FROM policies "+
-			"    JOIN policy_users pu ON policies.id = pu.policy_id "+
-			"    JOIN users on users.id = pu.user_id "+
-			"    AND ("+
+			"    LEFT JOIN policy_users pu ON policies.id = pu.policy_id "+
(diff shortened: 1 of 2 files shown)


application/models/policy.go
```diff
@@ -388,13 +388,13 @@ func scopeSearchPolicies(searchText string) pop.ScopeFunc {
 	return func(q *pop.Query) *pop.Query {
 		return q.Where("policies.id IN ("+
```
jason-jackson, line 389: Would this be better? (And does it still work?)
```suggestion
		return q.RawQuery("policies.id IN ("+
```

jason-jackson (approved):
Looks good. It would be nice if we could have used `q.LeftJoin()` in the scope function, but that didn't work (and I'm not certain it should work either given how pop works, meaning I think it's a feature, not a bug).

application/models/policy.go
briskt: No, it doesn't work. It appears to replace the entire query with only the text passed to RawQuery.

briskt: The function doc for RawQuery:
```
// RawQuery will override the query building feature of Pop and will use
// whatever query you want to execute against the `Connection`. You can continue
// to use the `?` argument syntax.
```

PR #607, opened 2023-11-21
fix nulls package use by always testing the Valid flag

briskt:
### Fixed
- Fixed code where `nulls` objects weren't tested for `Valid` before use.

[CVR-787](https://itse.youtrack.cloud/issue/CVR-787)

---

### Code Checklist
- [x] Run `gofmt`
- [x] Run `make swaggerspec`

### DIFF
--- a/application/actions/policies_test.go
+++ b/application/actions/policies_test.go
@@ -407,7 +407,7 @@ func (as *ActionSuite) Test_PoliciesUpdate() {
 			}
 			var policy api.Policy
 			as.NoError(json.Unmarshal([]byte(body), &policy))
-			as.Equal(*tt.update.HouseholdID, policy.HouseholdID)
+			as.Equal(*tt.update.HouseholdID, policy.HouseholdID.String)
 		})
 	}
 }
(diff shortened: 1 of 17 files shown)


application/actions/policies_test.go
```diff
@@ -407,7 +407,7 @@ func (as *ActionSuite) Test_PoliciesUpdate() {
 			}
 			var policy api.Policy
 			as.NoError(json.Unmarshal([]byte(body), &policy))
-			as.Equal(*tt.update.HouseholdID, policy.HouseholdID)
+			as.Equal(*tt.update.HouseholdID, policy.HouseholdID.String)
```
hobbitronics, line 410: I thought HouseholdID is validated at the struct level. Why does this differ from 620 of application/models/policy_test.go

hobbitronics (approved):
Still getting familiar with the nulls package so you may want to wait for another approval, but to me looks mostly good, just one question.

application/actions/policies_test.go
briskt: This isn't related to validation. It is just testing that the value is updated correctly. It's different than line 620 of policy_test.go because that line compares two values of type `nulls.String`. Here it is comparing a `nulls.String` with a `*string`.

briskt:
Looks like I added @forevermatt as a reviewer instead of @jason-jackson . Sorry about that.

PR #617, opened 2024-03-14
update Go to 1.22 and remove dependency on gobuffalo/buffalo image

briskt:
### Changed
- Updated to Go version 1.22
- Updated dependencies

### Removed
- Removed the gobuffalo/buffalo image dependency
- Removed npm from the Dockerfile-dev image


application/go.mod
```diff
@@ -1,11 +1,13 @@
 module github.com/silinternational/cover-api
 
-go 1.18
+go 1.21
+
+toolchain go1.22.1
```
jason-jackson, line 5: Any reason why not using go 1.22 here, just the toolchain?

jason-jackson (approved):

application/go.mod
briskt: The `toolchain` must be a new thing added by Go. I just overlooked changing line 3.

PR #619, opened 2024-04-05
respond to robots.txt with a global disallow

briskt:
### Added
- Respond to `/robots.txt` with "User-agent: * Disallow /"

---

### Code Checklist
- [x] Run `gofmt`
- [ ] Run `make swaggerspec`

### DIFF
--- a/application/actions/app.go
+++ b/application/actions/app.go
@@ -263,6 +263,11 @@ func App() *buffalo.App {
 		strikesGroup.PUT(idRegex, strikesUpdate)
 		strikesGroup.DELETE(idRegex, strikesDelete)
 
+		// robots
+		app.GET("/robots.txt", robots)
+		app.Middleware.Skip(AuthZ, robots)
+		app.Middleware.Skip(AuthN, robots)
+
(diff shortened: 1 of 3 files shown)


application/actions/app.go
```diff
@@ -263,6 +263,11 @@ func App() *buffalo.App {
 		strikesGroup.PUT(idRegex, strikesUpdate)
 		strikesGroup.DELETE(idRegex, strikesDelete)
 
+		// robots
+		app.GET("/robots.txt", robots)
+		app.Middleware.Skip(AuthZ, robots)
+		app.Middleware.Skip(AuthN, robots)
```
jason-jackson, line 269: Could also add this up by `homeHandler` and `statusHandler` (line 138-139), and add the Skip lines to lines 129, 130. But fine here too.

jason-jackson (approved):

application/actions/app.go
briskt: Huh, I guess there's not much reason to how these are ordered. I put it at the bottom since that's where "static" files are in CAM.

PR #625, opened 2024-05-23
set access token in session CVR-610

hobbitronics:
### Changed
- generate access token instead of getting from UI

### Removed
- ClientID

goes with ui PR: https://github.com/silinternational/cover-api/pull/625
---

### Code Checklist
- [ ] Run `gofmt`
- [x] Run `make swaggerspec`

### DIFF
--- a/application/actions/audits_test.go
+++ b/application/actions/audits_test.go
@@ -52,9 +52,7 @@ func (as *ActionSuite) Test_auditsRun() {
 	}
 	for _, tt := range tests {
 		as.T().Run(tt.name, func(t *testing.T) {
-			uat, err := tt.actor.CreateAccessToken(as.DB)
-			as.NoError(err)
-			as.Session.Set(AccessTokenSessionKey, uat.AccessToken)
+			as.SetAccessToken(tt.actor)
 			req := as.JSON(auditsPath)
 			req.Headers["content-type"] = domain.ContentJson
 			res := req.Post(tt.input)
--- a/application/actions/audits_test.go
+++ b/application/actions/audits_test.go
@@ -1,7 +1,6 @@
 package actions
 
 import (
-	"fmt"
 	"net/http"
 	"testing"
 	"time"
@@ -53,8 +52,10 @@ func (as *ActionSuite) Test_auditsRun() {
 	}
 	for _, tt := range tests {
 		as.T().Run(tt.name, func(t *testing.T) {
+			uat, err := tt.actor.CreateAccessToken(as.DB)
+			as.NoError(err)
+			as.Session.Set(AccessTokenSessionKey, uat.AccessToken)
 			req := as.JSON(auditsPath)
-			req.Headers["Authorization"] = fmt.Sprintf("Access %s", tt.actor.Email)
 			req.Headers["content-type"] = domain.ContentJson
 			res := req.Post(tt.input)
 			body := res.Body.Bytes()
--- a/application/actions/audits_test.go
+++ b/application/actions/audits_test.go
@@ -54,7 +54,7 @@ func (as *ActionSuite) Test_auditsRun() {
 	for _, tt := range tests {
 		as.T().Run(tt.name, func(t *testing.T) {
 			req := as.JSON(auditsPath)
-			req.Headers["Authorization"] = fmt.Sprintf("Bearer %s", tt.actor.Email)
+			req.Headers["Authorization"] = fmt.Sprintf("Access %s", tt.actor.Email)
 			req.Headers["content-type"] = domain.ContentJson
 			res := req.Post(tt.input)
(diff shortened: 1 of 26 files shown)


application/actions/auth.go
```diff
@@ -22,19 +22,15 @@ import (
...
+	AccessTokenSessionKey = "AccessToken"
 
 	// http param for an auth invite code
 	InviteCodeParam      = "invite"
 	InviteCodeSessionKey = "invite_code"
 
-	// logout http param for what is normally the bearer token
-	LogoutToken = "token"
+	// logout http param for what is normally the access token
+	LogoutToken = "access-token"
```
briskt, line 33: This shouldn't be needed now, I think. I'm not sure why we had the UI pass the token in the query string, but it seems like a bad idea now.

application/actions/auth_test.go
```diff
@@ -28,22 +28,22 @@ func (as *ActionSuite) Test_AuthLogin_Invite() {
 	}{
 		{
 			name: "Bad Invite Code",
-			queryParams: fmt.Sprintf("%s=123456&%s=badInviteCode",
-				ClientIDParam, InviteCodeParam),
+			queryParams: fmt.Sprintf("%s=badInviteCode",
+				InviteCodeParam),
```
briskt, line 32: This and other lines can be unwrapped
```suggestion
			queryParams: fmt.Sprintf("%s=badInviteCode", InviteCodeParam),
```

application/actions/audits_test.go
```diff
@@ -53,8 +52,10 @@ func (as *ActionSuite) Test_auditsRun() {
 	}
 	for _, tt := range tests {
 		as.T().Run(tt.name, func(t *testing.T) {
+			uat, err := tt.actor.CreateAccessToken(as.DB)
+			as.NoError(err)
+			as.Session.Set(AccessTokenSessionKey, uat.AccessToken)
```
briskt, line 57: Due to how many times this is repeated, a new helper might be worth it:
```go
func (as *ActionSuite) SetAccessToken(actor models.User) {
    uat, err := actor.CreateAccessToken(as.DB)
    as.NoError(err)
    as.Session.Set(AccessTokenSessionKey, uat.AccessToken)
}

File: application/api/errorcodes.go [hunk 4]
```diff
@@ -56,9 +56,7 @@ const (
 	ErrorInviteExpired            = ErrorKey("ErrorInviteExpired")
 	ErrorLoadingAuthProvider      = ErrorKey("ErrorLoadingAuthProvider")
 	ErrorMissingAuthEmail         = ErrorKey("ErrorMissingAuthEmail")
-	ErrorMissingClientID          = ErrorKey("ErrorMissingClientID")
 	ErrorMissingLogoutToken       = ErrorKey("ErrorMissingLogoutToken")
```
briskt, line 59: I think this can go away
```suggestion
```

application/domain/domain.go
```diff
@@ -481,6 +462,14 @@ func IsProduction() bool {
...
 }
 
+func IsDevelopment() bool {
+	return Env.GoEnv == EnvDevelopment
+}
+
+func IsTest() bool {
+	return Env.GoEnv == EnvTest
+}
+
```
briskt, line 472: Are these used?

application/models/useraccesstoken.go
```diff
@@ -149,16 +149,16 @@ func (u *UserAccessToken) Update(tx *pop.Connection) error {
 }
 
 // InitAccessToken prepares a new value for the AccessToken field and the ExpiresAt field.
-func InitAccessToken(clientID string) UserAccessToken {
+func InitAccessToken() UserAccessToken {
 	token, _ := getRandomToken() // The init() function would have made sure there was no error
 
 	if domain.Env.GoEnv == domain.EnvDevelopment {
-		fmt.Printf("\n\nClientID+token: %s%s\n", clientID, token)
+		fmt.Printf("\n\ntoken: %s\n", token)
```
briskt, line 156: I think this can (should?) be removed. It may have been added because of the difficulty in debugging tokens with and without the client Id added. Even if it remains, it's better to make it a `log.Debug` call instead of `fmt.Printf`.

briskt (approved):
Looks really good! I added a few minor suggestions.

application/actions/audits_test.go
hobbitronics: Do we have a place for helpers specifically for tests, or can I just put it in domain.go?

briskt: Can't go in domain.go because that's in a different package. It should be in the actions package, and it should be in a file that ends in `_test.go` so it's not available to the application itself. More specifically, this one should be in actions_test.go.

PR #629, opened 2024-06-04
allow deleting dependent with Inactive item

hobbitronics:
### Fixed
- Allow deleting a dependent with an Inactive item, since the item can't be updated
[CVR-813](https://itse.youtrack.cloud/issue/CVR-813)

### Code Checklist
- [ ] Run `gofmt`
- [ ] Run `make swaggerspec`


briskt (approved):

hobbitronics:
> Was there not a test that had to be updated?

This didn't cause any tests to fail, but I've since added an Inactive Item to deletableDep to cover this change.

PR #634, opened 2024-06-10
add premium rate to items CVR-692

hobbitronics:
### Added
- Item.PremiumRate

[CVR-692](https://itse.youtrack.cloud/issue/CVR-692) include per-item premium rate in API response

---

### Code Checklist
- [ ] Run `gofmt`
- [ ] Run `make swaggerspec`

### DIFF
--- a/application/models/item.go
+++ b/application/models/item.go
@@ -934,7 +934,7 @@ func (i *Item) CalculateAnnualPremium(tx *pop.Connection) api.Currency {
 	factor := i.GetPremiumFactor(tx)
 	premium := int(math.Round(float64(i.CoverageAmount) * factor))
 
-	return api.Currency(domain.Max(premium, i.Category.MinimumPremium))
+	return api.Currency(max(premium, i.Category.MinimumPremium))
 }
 
 // GetPremiumFactor returns the premium rate for the item's category
--- a/application/models/item.go
+++ b/application/models/item.go
@@ -69,6 +69,7 @@ type Item struct {
 	Category     ItemCategory `belongs_to:"item_categories" validate:"-"`
 	RiskCategory RiskCategory `belongs_to:"risk_categories" validate:"-"`
 	Policy       Policy       `belongs_to:"policies" validate:"-"`
+	PremiumRate  float64      `db:"-"`
 }
 
 // Validate gets run every time you call pop.ValidateAndSave, pop.ValidateAndCreate, or pop.ValidateAndUpdate
@@ -820,6 +821,7 @@ func (i *Item) ConvertToAPI(tx *pop.Connection) api.Item {
 		AnnualPremium:         i.CalculateAnnualPremium(tx),
 		MonthlyPremium:        i.CalculateMonthlyPremium(tx),
 		ProratedAnnualPremium: i.CalculateProratedPremium(tx, time.Now().UTC()),
+		PremiumRate:           math.Round(i.GetPremiumFactor(tx)*100) / 100,
 		CanBeDeleted:          i.canBeDeleted(tx),
 		CanBeUpdated:          !i.hasOpenClaim(tx),
 		CreatedAt:             i.CreatedAt,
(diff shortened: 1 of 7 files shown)


briskt:
What is the purpose of this? Is it a step towards full implementation of CVR-692, per-item premium customized rates? As it is, it might raise confusion about such a feature being available but is not acually.

application/models/item.go
```diff
@@ -928,14 +930,21 @@ func (i *Item) CalculateBillingPremium(tx *pop.Connection) api.Currency {
 // CalculateAnnualPremium returns the rounded product of the item's CoverageAmount and the category's
 // PremiumFactor, with the category's minimum premium applied
 func (i *Item) CalculateAnnualPremium(tx *pop.Connection) api.Currency {
+	i.LoadCategory(tx, false)
+	factor := i.GetPremiumFactor(tx)
+	premium := int(math.Round(float64(i.CoverageAmount) * factor))
+
+	return api.Currency(domain.Max(premium, i.Category.MinimumPremium))
```
jason-jackson, line 937: Let's ditch the domain.Max use as we are on go 1.22 now, so we can use the builtin `max` function instead.

```suggestion
	return api.Currency(max(premium, i.Category.MinimumPremium))
```

application/models/item.go
```diff
@@ -69,6 +69,7 @@ type Item struct {
 	Category     ItemCategory `belongs_to:"item_categories" validate:"-"`
 	RiskCategory RiskCategory `belongs_to:"risk_categories" validate:"-"`
 	Policy       Policy       `belongs_to:"policies" validate:"-"`
+	PremiumRate  float64      `db:"-"`
```
jason-jackson, line 72: I'm not seeing this being used anywhere except in a test. I don't think it's needed and could be deleted
```suggestion
```

application/models/item_test.go
```diff
@@ -1504,6 +1504,7 @@ func (ms *ModelSuite) TestItem_ConvertToAPI() {
 	fixtures := CreateItemFixtures(ms.DB, FixturesConfig{DependentsPerPolicy: 1})
 	item := fixtures.Items[0]
 	item.CoverageEndDate = nulls.NewTime(time.Now().Add(domain.DurationDay * 365))
+	item.PremiumRate = .02
```
jason-jackson, line 1507: item.PremiumRate is never used outside of this test and can be deleted

```suggestion
```

application/models/item_test.go
```diff
@@ -1541,6 +1542,7 @@ func (ms *ModelSuite) TestItem_ConvertToAPI() {
 		"AccountablePerson Country is not correct")
 	ms.True(got.CanBeUpdated, "CanBeUpdated is not correct")
 	ms.True(got.CanBeDeleted, "CanBeDeleted is not correct")
+	ms.Equal(item.PremiumRate, got.PremiumRate, "PremiumRate is not correct")
```
jason-jackson, line 1545: item.PremiumRate is never used outside of this test and can be deleted

```suggestion
```

application/models/item.go
jason-jackson: Note: This probably is what @briskt was talking about. I think we might need a bit more discussion on the ticket before this is fully implemented. Like what should we show when premiums are changing? Do we show old value? New value? Both values? If it's just the new value, we don't need to store it, otherwise we'd need to store it (and update it after next billing cycle).

hobbitronics: I think it's needed for the LedgerEntry. I am still working on that though.

hobbitronics:
> What is the purpose of this? Is it a step towards full implementation of CVR-692, per-item premium customized rates? As it is, it might raise confusion about such a feature being available but is not acually.

After looking more closely at the parent/ui issue I see that "premium rate per item" is ultimately for the Report on a Policy. I am still actively working on that this isn't ready by itself. More discussion would would be helpful as @jason-jackson suggests.

PR #635, opened 2024-06-12
add TestDestroyIfExpired CVR-820

hobbitronics:
### Added
- add TestDestroyIfExpired to policyuserinvite_test.go

[CVR-820](https://itse.youtrack.cloud/issue/CVR-820)
---

### Code Checklist
- [ ] Run `gofmt`
- [ ] Run `make swaggerspec`

### DIFF
--- a/application/models/policyuserinvite_test.go
+++ b/application/models/policyuserinvite_test.go
@@ -45,19 +45,20 @@ func (ms *ModelSuite) TestDestroyIfExpired() {
 	policies := policyFixtures.Policies
 
 	createUniqueInvite := func(createdAt time.Time) PolicyUserInvite {
+		randomStr := randStr(5)
 		return PolicyUserInvite{
 			ID:           domain.GetUUID(),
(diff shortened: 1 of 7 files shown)


application/models/policyuserinvite_test.go
```diff
@@ -32,3 +34,80 @@ func (ms *ModelSuite) TestPolicyUserInvite_ConvertToAPI() {
...
+			InviterEmail: "test@example.org",
+			CreatedAt:    createdAt,
+		}
+	}
+
+	countInvites := func() int {
+		count, err := tx.Count(&PolicyUserInvites{})
+		ms.NoError(err)
+		return count
+	}
```
briskt, line 63: It may be a problem at some point that this test requires that no invites exist other than the ones created by this test. Perhaps this could be refined to count only invites for the policy used by this test.

briskt (approved):
Good idea to add this test. I do suggest making this test able to work even with other invites present in the database.

briskt (approved):

PR #637, opened 2024-06-12
add invite expired event CVR-820

hobbitronics:
### Added
- add invite expired event

[CVR-820](https://itse.youtrack.cloud/issue/CVR-820)
---

### Code Checklist
- [ ] Run `gofmt`
- [x] Run `make swaggerspec`

### DIFF
--- a/application/listeners/policy_test.go
+++ b/application/listeners/policy_test.go
@@ -2,6 +2,7 @@ package listeners
 
 import (
 	"testing"
+	"time"
 
 	"github.com/gobuffalo/events"
 	"github.com/gofrs/uuid"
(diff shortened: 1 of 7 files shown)


application/models/policyuserinvite_test.go
```diff
@@ -74,19 +64,19 @@ func (ms *ModelSuite) TestDestroyIfExpired() {
...
 			name:   "Invite expired",
-			invite: createUniqueInvite(cutoff.Add(-time.Hour)),
+			invite: CreateUniqueInvite(cutoff.Add(-time.Hour), pID2),
 			expectedErr: api.NewAppError(
 				errors.New("attempt to use expired invite, ID: "),
 				api.ErrorInviteExpired,
 				api.CategoryForbidden,
 			),
-			expectedCount: 0,
+			expectedCount: 1,
```
hobbitronics, line 79: I am thinking there is a delay, but the event will cause it to be 0 eventually.

briskt (approved):
looks great!

application/models/policyuserinvite_test.go
briskt: I wonder if adding a small delay to the test will make it work. Probably not, because I think the event listener isn't running during the model test. But there is a way to confirm the event was fired. Check out `TestItem_Approve`.

hobbitronics: Doing that didn't seem to change anything.

application/listeners/policy_test.go
```diff
@@ -64,3 +65,56 @@ func (ts *TestSuite) Test_PolicyUserInviteSend() {
 		})
 	}
 }
+
+func (ts *TestSuite) Test_PolicyUserInviteExpired() {
```
briskt, line 69: The SonarQube analysis highlight this function name. While their regular expression doesn't work for us, this name does diverge from our norm. For consistency, it should be `TestPolicyUser_InviteExpired`. The underscore separates the struct name from the test name. In general, function names shouldn't have underscores. It could be worth reconsidering our practice of adding underscores in test names.

PR #640, opened 2024-06-14
return ErrorInviteExpired

hobbitronics:
### Changed
- return ErrorInviteExpired

---

### Code Checklist
- [ ] Unit tests created or updated
- [x] Run `gofmt`
- [ ] Run `make swaggerspec`

### DIFF
--- a/application/actions/auth.go
+++ b/application/actions/auth.go
@@ -145,7 +145,7 @@ func validateInviteOnLogin(c buffalo.Context, inviteCode string) *api.AppError {
 	}
 
 	if err := invite.DestroyIfExpired(tx); err != nil {
-		return &appErr
+		return err.(*api.AppError)
 	}
 
 	c.Session().Set(InviteCodeSessionKey, inviteCode)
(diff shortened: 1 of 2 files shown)


application/actions/auth.go
```diff
@@ -145,7 +145,7 @@ func validateInviteOnLogin(c buffalo.Context, inviteCode string) *api.AppError {
 	}
 
 	if err := invite.DestroyIfExpired(tx); err != nil {
-		return &appErr
+		return err.(*api.AppError)
```
briskt, line 148: This is unusual. If this function would have `error` is it's return type, all you'd need is `return err`. Not expecting you to fix that, just calling it out.

briskt (approved):

application/actions/auth.go
hobbitronics: Do you mean validateInviteOnLogin by this?

briskt: Yes

PR #649, opened 2024-07-20
create Admin role -- same as Steward but without the notifications

briskt:
### Added
- Created an Admin role, the same as Steward but without the notifications.

---

### Code Checklist
- [ ] Unit tests created or updated
- [x] Run `gofmt`
- [ ] Run `make swaggerspec`


hobbitronics (commented):
Surprisingly little code leaves me with some questions. Do we need any tests for this role? I suppose the Steward tests should suffice?

~~Also we could set the id range for an Admin like we do for AppRoleSteward and AppRoleSignator in user.go line 149~~ I somehow missed this, but see you did.

hobbitronics (approved):

briskt:
> Surprisingly little code leaves me with some questions. Do we need any tests for this role? I suppose the Steward tests should suffice?
> 

The only test I can think of is a test that would prove that no emails are sent to an Admin. I don't know if that's very important, though.

jason-jackson (approved):

PR #651, opened 2024-07-29
run golangci-lint

briskt:
### Added
- Run golangci-lint in GitHub Actions

### Fixed
- Fixed issues identified by golangci-lint


application/auth/saml/provider.go
```diff
@@ -254,8 +245,7 @@ func decodeKey(key, expectedType string) ([]byte, error) {
 		return block.Bytes, nil
 	}
 
-	var bytes []byte
-	bytes = make([]byte, base64.StdEncoding.DecodedLen(len(key)))
+	bytes := make([]byte, base64.StdEncoding.DecodedLen(len(key)))
```
hobbitronics, line 248: I understand this will give bytes the length it will have after decoding `key`. What was the linter complaining about not specifying the length?

application/domain/domain.go
```diff
@@ -338,7 +338,6 @@ func IsStringInSlice(needle string, haystack []string) bool {
 }
 
 func RandomString(n int, includeLetters string) string {
-	rand.Seed(time.Now().UnixNano())
```
hobbitronics, line 341: What's the problem or lack of need here?

.github/workflows/test-and-deploy.yml
```diff
@@ -19,6 +19,12 @@ jobs:
       - uses: actions/checkout@v4
       - name: Run unit tests
         run: docker-compose -f actions-services.yml run --rm app ./run-tests.sh
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v6
+        with:
+          version: latest
```
hobbitronics, line 25: Should this match the version we are on?

hobbitronics (approved):
Nice cleanup! A few questions

application/auth/saml/provider.go
briskt: It was just too verbose.
```
auth/saml/provider.go:270:2: S1021: should merge variable declaration with assignment on next line (gosimple)
        var bytes []byte
        ^
```

application/domain/domain.go
briskt: If I understand correctly, it's just unnecessary because the library provides a random seed by default.
```
domain/domain.go:341:2: SA1019: rand.Seed has been deprecated since Go 1.20 and an
alternative has been available since Go 1.0: As of Go 1.20 there is no reason to call 
Seed with a random value. Programs that call Seed with a known value to get a specific 
sequence of results should use New(NewSource(seed)) to obtain a local random generator. 
(staticcheck)
        rand.Seed(time.Now().UnixNano())
        ^
```

application/auth/saml/provider.go
hobbitronics: I thought the former was less verbose. Maybe I don't know what it means.

